PR #1303, opened 2024-02-11
CC-711 track concept progress

libmartinito:
**Checklist**:

## Summary by CodeRabbit

- **New Features**
	- Introduced a progress tracking feature for concepts, including a visual progress bar on concept cards.
	- Added the ability to create and update concept engagements based on user interactions.
- **Enhancements**
	- Improved authentication checks before fetching concept-related data, ensuring data is fetched based on the user's authentication status.
- **Bug Fixes**
	- Removed unnecessary update hook in `concept.hbs` to enhance performance.
- **Tests**
	- Added new acceptance tests for progress tracking and concept completion.
- **Documentation**
	- Updated templates and components documentation to reflect changes in progress tracking and authentication handling.
- **Chores**
	- Added routes and factories in MirageJS for concept engagements handling.


rohitpaulk (dismissed):
Added one round of comments, will take a look again once these are addressed. 

Can we also have the backend PR up for review? That way I can test this on the vercel preview URL and see what it feels like

app/components/concept-page/content.ts
```diff
@@ -28,16 +38,6 @@ export default class ContentComponent extends Component<Signature> {
     element.scrollIntoView({ behavior: 'smooth', block: 'nearest' });
   }
 
-  @action
-  handleConceptDidUpdate() {
-    this.currentProgressPercentage = 0;
-  }
```
libmartinito, line 34: can you confirm if removing this works? i feel like we don't want to reset the progress to 0 on concept update

rohitpaulk: @libmartinito this was intended to handle cases where you navigate from one concept to another.

If you go to a concept that is part of a collection and there's a next concept in the collection, here's what should happen:

https://github.com/codecrafters-io/frontend/assets/3893573/213c56e9-283c-43d1-966c-b2f859c86305

Without this hook, previously, the page would've update to the new concept but the progress would remain the same (100%). 

Should be safe to remove now since `conceptEngagement` is fetched in the route and passed down - so when `Concept` changes, `conceptEngagement` will change too.

app/components/concept.ts
```diff
@@ -97,6 +106,28 @@ export default class ConceptComponent extends Component<Signature> {
...
+        break;
+      } else if (nextSum === completedBlocksCount) {
+        currentBlockGroupIndex = i + 1;
+        break;
+      }
+
+      sum = nextSum;
+    }
+
+    return currentBlockGroupIndex;
```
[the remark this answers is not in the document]
rohitpaulk: Was answered over call - this is computed from progress %

app/components/concept.ts
```diff
@@ -168,4 +191,11 @@ export default class ConceptComponent extends Component<Signature> {
       window.history.replaceState({}, '', `${window.location.pathname}?${urlParams.toString()}`);
     }
   }
+
+  updateProgressPercentage(progressPercentage: number) {
+    if (progressPercentage > this.args.latestConceptEngagement.currentProgressPercentage) {
```
rohitpaulk, line 196: @libmartinito do you think we need this check? If a user clicks on "step back" on a concept a couple of times, maybe they expect to be taken back to where they were? Would everything work fine if we removed this check and allowed `progressPercentage` to decrease if a user clicks on "step back"?

app/components/concept.ts
```diff
@@ -157,7 +180,7 @@ export default class ConceptComponent extends Component<Signature> {
 
   updateLastRevealedBlockGroupIndex(newBlockGroupIndex: number) {
     this.lastRevealedBlockGroupIndex = newBlockGroupIndex;
-    this.args.onProgressPercentageChange(this.progressPercentage);
+    this.updateProgressPercentage(this.progressPercentage);
```
rohitpaulk, line 183: It's a bit weird that we're passing in `this.progressPercentage` now, since the `updateProgressPercentage` method already has access to `this.progressPercentage`. 

It'a also confusing that there's `conceptEngagement.progressPercentage` and `progressPercentage`, and it's not super clear how they're different. 

How about this: 

- Remove `this.progressPercentage` from this method call, the method already has access to `this.progressPercentage`
- Name the method `updateConceptEngagement` instead of `updateProgressPercentage`
- Rename `this.progressPercentage` to `this.computedProgressPercentage`, that way it's clear that we call this when we want to "compute", and it's actually stored on `conceptEngagement.progressPercentage`.

tests/acceptance/concepts-test.js
```diff
@@ -22,6 +22,10 @@ module('Acceptance | concepts-test', function (hooks) {
   setupAnimationTest(hooks);
   setupWindowMock(hooks);
 
+  hooks.beforeEach(function () {
+    time.runAtSpeed(100);
```
rohitpaulk, line 26: @libmartinito is this still needed?

app/routes/concept.js
```diff
@@ -16,6 +17,24 @@ export default class ConceptRoute extends BaseRoute {
...
+      .sortBy('createdAt')
+      .reverse()[0];
+
+    if (!latestConceptEngagement) {
+      return await this.store
+        .createRecord('concept-engagement', {
+          concept,
+          user: this.authenticator.currentUser,
+        })
+        .save();
```
rohitpaulk, line 32: I don't think we need to `save()` here. I think it'll implicitly get saved when a user ends up clicking on the first block, no? Similar to repositories. We create a repository record when a user starts, but we don't call `save()` until the user chooses their language.

I'm not sure we'd want to count a "concept engagement" if a user only loads the page but does nothing with it.

app/routes/concept.js
```diff
@@ -16,6 +17,24 @@ export default class ConceptRoute extends BaseRoute {
...
+    const latestConceptEngagement = this.authenticator.currentUser?.conceptEngagements
+      .filter((engagement) => engagement.concept.slug === concept.slug)
+      .sortBy('createdAt')
+      .reverse()[0];
+
+    if (!latestConceptEngagement) {
+      return await this.store
+        .createRecord('concept-engagement', {
+          concept,
+          user: this.authenticator.currentUser,
```
rohitpaulk, line 30: What if the user isn't authenticated?

rohitpaulk (dismissed):
Comments added. Primary problem would be that this breaks concepts for unauthenticated users - let's fix this + add a test so that we're covered in the future

app/components/concept-card/progress-bar.hbs
```diff
@@ -0,0 +1,23 @@
...
+      {{/if}}
+    </div>
+  </div>
+
+  <div
+    class="text-xs {{if (eq @latestConceptEngagement.currentProgressPercentage 100) 'text-teal-500' 'text-blue-500'}} mr-3"
+    data-test-concept-card-progress-text
+  >
+    <b class="font-bold">{{@latestConceptEngagement.currentProgressPercentage}}</b>
+    % complete
```
rohitpaulk, line 21:
```suggestion
    <b class="font-bold">{{@latestConceptEngagement.currentProgressPercentage}}%</b>
    complete
```

(minor design change)

app/components/concept.ts
```diff
@@ -93,38 +99,66 @@ export default class ConceptComponent extends Component<Signature> {
...
+      } else if (traversedBlocksCount === completedBlocksCount) {
+        currentBlockGroupIndex = i + 1;
+        break;
+      }
+    }
+
+    return currentBlockGroupIndex;
+  }
+
+  enqueueConceptEngagementUpdate = task({ enqueue: true }, async () => {
```
rohitpaulk, line 130: @libmartinito how do we choose between `enqueue`, and `keepLatest` here? What are the downsides/upsides to using each?

rohitpaulk:
There seems to be a bug here - clicking on "Continue" causes the progress bar to flash and then not appear: 


https://github.com/codecrafters-io/frontend/assets/3893573/fdf4e7e9-1545-4eed-8b24-9bb91a52a205

rohitpaulk:
Ah, ignore - looks like there was a WIP core PR that needed to be merged. Just merged, will test soon

app/routes/concept.js
```diff
@@ -25,10 +42,20 @@ export default class ConceptRoute extends BaseRoute {
...
       .sort((a, b) => a.slug.localeCompare(b.slug));
 
+    if (this.authenticator.isAuthenticated) {
+      await this.store.findAll('concept-engagement', {
+        include: 'concept,user',
+      });
+    }
+
+    let latestConceptEngagement;
+    latestConceptEngagement = await this.findOrCreateConceptEngagement(concept);
```
rohitpaulk, line 52: @libmartinito can't we just use `const xyz = ...` here?

app/models/concept.ts
```diff
@@ -13,8 +14,9 @@ export type BlockJSON = {
 };
 
 export default class ConceptModel extends Model {
-  @belongsTo('user', { async: false, inverse: null }) declare author: UserModel;
+  @belongsTo('user', { async: true, inverse: null }) declare author: UserModel;
```
rohitpaulk, line 17: @libmartinito we don't use `async: true` relationships as a rule since they can fire requests at unexpected times. Can you elaborate on why we need this?

app/components/concept.ts
```diff
@@ -150,14 +184,21 @@ export default class ConceptComponent extends Component<Signature> {
...
       this.updateLastRevealedBlockGroupIndex(this.currentBlockGroupIndex - 1);
+      this.enqueueConceptEngagementUpdate.perform();
     }
 
     // TODO: Add analytics event?
   }
 
+  @action
+  handleWillDestroyContainer() {
+    if (!this.authenticator.isAuthenticated && config.environment !== 'test') {
```
rohitpaulk, line 195: @libmartinito what was the deal with this in the test environment? what kind of test failure do we see if this test environment filter is removed?

rohitpaulk (changes requested):
Added questions!

app/models/concept.ts
libmartinito: seems to be fixed. don't need anymore

app/components/concept.ts
libmartinito: tried using unloadRecord. didn't work. will make an issue after this gets merged so i could reference code. will update here after making issue

PR #1328, opened 2024-02-15
CC-1011 add share button to concepts

libmartinito:
**Checklist**:

## Summary by CodeRabbit

- **New Features**
	- Introduced a new section for sharing concepts with friends, complete with a gift icon and a copyable code snippet feature.
- **Style**
	- Adjusted padding for the concept section for improved layout.
	- Enhanced the share concept container with a border, rounded corners, shadow, and padding for better visual appeal.
- **Tests**
	- Added assertions and tests to ensure proper tracking of analytics events related to concept sharing.

### DIFF
--- a/app/components/concept-page/content.hbs
+++ b/app/components/concept-page/content.hbs
@@ -53,7 +53,7 @@
                   Learnt something new? Share it with a friend.
                 </span>
 
-                <CopyableCode @code={{this.conceptUrl}} @onCopyButtonClick={{this.handleCopyButtonClick}} />
+                <CopyableCode @code={{@concept.url}} @onCopyButtonClick={{this.handleCopyButtonClick}} />
               </div>
             </div>
 
--- a/app/components/concept-page/content.hbs
+++ b/app/components/concept-page/content.hbs
@@ -34,7 +34,7 @@
               You've completed this concept!
             </div>
 
-            <FeedbackButton @source="concept_completed" class="mr-4" as |dd|>
+            <FeedbackButton @source="concept_completed" class="mr-4 mb-16" as |dd|>
               <div
                 role="button"
                 class="rounded border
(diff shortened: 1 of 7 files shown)


tests/acceptance/concepts-test.js
```diff
@@ -164,6 +165,55 @@ module('Acceptance | concepts-test', function (hooks) {
...
+    await conceptPage.questionCards[3].clickOnShowExplanationButton();
+    await conceptPage.clickOnContinueButton();
+    await conceptPage.clickOnContinueButton();
+    await conceptPage.clickOnContinueButton();
+    await conceptPage.clickOnContinueButton();
+    await conceptPage.clickOnContinueButton();
+
+    await conceptPage.shareConceptContainer.clickOnCopyButton();
+
+    const store = this.owner.lookup('service:store');
```
rohitpaulk, line 210: Can we assert via `mirage` instead of using `store`? That way the API part of things is tested too. 

If we're doing this same thing elsewhere, let's try to change that too please. We don't want to end up with errors where something is present in the store but wasn't saved to the API (for example, if we forgot to add `save()` to the record, think this test would still pass since it accesses the store directly?)

app/components/concept-page/content.hbs
```diff
@@ -60,6 +60,17 @@
               </div>
             {{/animated-if}}
           </div>
+
+          <div class="py-12 border-t" {{did-insert this.handleShareConceptContainerInserted}} data-test-share-concept-container>
+            <div class="flex flex-col items-start gap-4 border rounded shadow-sm p-4">
+              <span class="flex gap-2 items-baseline">
```
rohitpaulk, line 66: Alignment here looks off: 

Screenshot 2024-02-15 at 21 09 59

rohitpaulk: Let's do this instead?

```suggestion
              <span class="flex gap-2 items-center text-gray-700 text-sm">
```

Screenshot 2024-02-15 at 21 12 18

Screenshot 2024-02-15 at 21 12 23

app/components/concept-page/content.ts
```diff
@@ -33,11 +41,23 @@ export default class ContentComponent extends Component<Signature> {
...
+  }
+
   @action
   handleProgressPercentageChanged(progressPercentage: number) {
     this.currentProgressPercentage = progressPercentage;
   }
 
+  @action
+  handleShareConceptContainerInserted(element: HTMLElement) {
+    element.scrollIntoView({ behavior: 'smooth', block: 'nearest' });
```
rohitpaulk, line 58: Looks like we'll be inserting three separate elements and calling `scrollIntoView` on them - completion container, next concept and share concept.. Can we just scroll one instead? Maybe we could place both "next concept" and "share concept" within "completion container", and only keep the scroll handler that's present there? That way the whole block will be scrolled into view.

app/components/concept-page/content.hbs
```diff
@@ -60,6 +60,17 @@
               </div>
             {{/animated-if}}
           </div>
+
+          <div class="py-12 border-t" {{did-insert this.handleShareConceptContainerInserted}} data-test-share-concept-container>
```
rohitpaulk, line 64: Let's place this above `nextConcept` please, feels more structured/coherent to have the order be "share feedback on current concept", "share current concept", "move on to next concept". 

And we don't need `animated-if` for `nextConcept` above - that value isn't going to change while the user is viewing a page (it's loaded when the concept page is loaded). A plain `#if` should do I think

app/components/concept-page/content.ts
```diff
@@ -13,8 +15,14 @@ interface Signature {
 }
 
 export default class ContentComponent extends Component<Signature> {
+  @service declare analyticsEventTracker: AnalyticsEventTrackerService;
+
   @tracked currentProgressPercentage = 0;
 
+  get conceptUrl() {
+    return `https://app.codecrafters.io/concepts/${this.args.concept.slug}`;
```
rohitpaulk, line 23: Let's place this in `Concept#url` instead? That way we can re-use it in other places if needed.

rohitpaulk (approved):

PR #1349, opened 2024-02-19
CC-1021 add 'invite friend' button below leaderboard

libmartinito:
**Checklist**:

### DIFF
--- a/app/components/invite-button.ts
+++ b/app/components/invite-button.ts
@@ -1,4 +1,6 @@
+import type AuthenticatorService from 'codecrafters-frontend/services/authenticator';
 import Component from '@glimmer/component';
+import { inject as service } from '@ember/service';
 
 export interface Signature {
   Element: HTMLDivElement;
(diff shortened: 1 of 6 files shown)


libmartinito:
failing test is due to timeout. it seems to be not related to changes. already tried rerunning a few times. will rerun tom

app/components/invite-button.ts
```diff
@@ -0,0 +1,23 @@
+import type AuthenticatorService from 'codecrafters-frontend/services/authenticator';
+import Component from '@glimmer/component';
+import { inject as service } from '@ember/service';
+
+export interface Signature {
+  Element: HTMLDivElement;
+
+  Args: {
+    isCollapsed: boolean;
+    onClick: () => void;
```
rohitpaulk, line 10: This doesn't need to be accepted as an arg, you can directly use `{{on "click" <whatever>}}` from the caller

PR #1386, opened 2024-02-27
CC-945: Update ordering of courses in user profile.

ryan-gang:
**Checklist**:

## Summary by CodeRabbit

- **New Features**
	- Introduced a `CourseProgressListComponent` to organize and display user course participations by course, including sorting by completion status and last submission date.
	- Added a `CourseParticipationModel` to represent a user's participation in a course, including completion and submission dates.

- **Tests**
	- Updated test descriptions and logic to include proper ordering of courses based on completion and submission dates.
	- Added new page object properties for course name and description in course progress list items.
	- Enhanced user profile view tests with assertions for course ordering and added new collections for targeting course elements.


app/components/user-page/course-progress-list.hbs
```diff
@@ -5,6 +5,7 @@
 
   <div class="grid grid-cols-1 gap-3 lg:grid-cols-2">
     {{#each this.courseParticipationGroups as |courseParticipationGroup|}}
+      {{! @glint-ignore }}
```
rohitpaulk, line 8: @ryan-gang let's do `@glint-expect-error` instead of `@glint-ignore`. `@glint-ignore` won't complain if the error isn't present anymore.

More on this here: https://typed-ember.gitbook.io/glint/using-glint/directives

app/models/course-participation.ts
```diff
@@ -0,0 +1,21 @@
...
+import type LanguageModel from './language';
+
+export default class CourseParticipationModel extends Model {
+  @belongsTo('course', { async: false, inverse: null }) declare course: CourseModel;
+  @belongsTo('course-stage', { async: false, inverse: null }) declare currentStage: CourseStageModel;
+  @belongsTo('user', { async: false, inverse: 'courseParticipations' }) declare user: UserModel;
+  @belongsTo('language', { async: false, inverse: null }) declare language: LanguageModel;
+
+  @attr('date') declare completedAt: Date;
+  @attr('date') declare lastActivityAt: Date;
```
rohitpaulk, line 14: Let's remove this, since it isn't being returned by the backend. Same for `started_at`

app/models/course-participation.ts
```diff
@@ -0,0 +1,21 @@
...
+import type UserModel from './user';
+import type LanguageModel from './language';
+
+export default class CourseParticipationModel extends Model {
+  @belongsTo('course', { async: false, inverse: null }) declare course: CourseModel;
+  @belongsTo('course-stage', { async: false, inverse: null }) declare currentStage: CourseStageModel;
+  @belongsTo('user', { async: false, inverse: 'courseParticipations' }) declare user: UserModel;
+  @belongsTo('language', { async: false, inverse: null }) declare language: LanguageModel;
+
+  @attr('date') declare completedAt: Date;
```
rohitpaulk, line 13: This is nullable, so the type should be `Date | null`

app/models/course-participation.ts
```diff
@@ -0,0 +1,21 @@
...
+  @attr('date') declare lastActivityAt: Date;
+  @attr('date') declare startedAt: Date;
+  @attr('date') declare lastSubmissionAt: Date;
```
rohitpaulk, line 16: Same for this - type should be `Date | null`

tests/acceptance/view-user-profile-test.js
```diff
@@ -63,6 +81,11 @@ module('Acceptance | view-user-profile', function (hooks) {
     });
 
     await userPage.visit({ username: 'rohitpaulk' });
+    assert.strictEqual(userPage.courseLabels[0].text, 'Build your own Docker');
+    // Completed course will be listed first.
```
rohitpaulk, line 85: Keep comments above the lines that they apply to - isn't Docker the completed one here?

tests/acceptance/view-user-profile-test.js
```diff
@@ -63,6 +81,11 @@ module('Acceptance | view-user-profile', function (hooks) {
...
+    assert.strictEqual(userPage.courseLabels[1].text, 'Build your own grep');
+    assert.strictEqual(userPage.courseLabels[2].text, 'Build your own Redis');
+    // Then ordering is done on latest submission, earlier comes first.
```
rohitpaulk, line 88: Same as above - these comments are really confusing, it feels like you're commenting on the `percySnapshot` line below

tests/pages/user-page.js
```diff
@@ -11,5 +11,7 @@ export default createPage({
     hover: triggerable('mouseenter'),
   },
 
+  courseLabels: collection('[data-test-course-name]'),
```
rohitpaulk, line 14: Let's try to map these to component names. In this case, it'd be `courseProgressListItems`. That way, in the future if we want to test other properties (like progress bar width etc.), it's easy to add those properties to this page object.

app/components/user-page/course-progress-list.ts
```diff
@@ -0,0 +1,43 @@
...
+      } else {
+        // If both groups have completed participations or both don't, sort by the lastSubmissionAt of the last participation in the group
+        const lastSubmissionA = new Date(groupA[groupA.length - 1]!.lastSubmissionAt);
+        const lastSubmissionB = new Date(groupB[groupB.length - 1]!.lastSubmissionAt);
+
+        return lastSubmissionA.getTime() - lastSubmissionB.getTime();
+      }
+    });
+
+    console.log(sortedParticipationsGroupedByCourse);
```
rohitpaulk, line 39: @ryan-gang why do you have a console.log here - please self-review before submitting for review, things like console.log and commented code are very basic things that we should never miss

app/components/user-page/course-progress-list.ts
```diff
@@ -0,0 +1,43 @@
...
+    user: UserModel;
+  };
+};
+
+export default class CourseProgressListComponent extends Component<Signature> {
+  get courseParticipationGroups() {
+    const groupedCourseParticipations: CourseParticipationModel[][] = Object.values(groupBy(this.args.user.courseParticipations, 'course'));
+
+    const sortedParticipationsGroupedByCourse = groupedCourseParticipations.map((group) =>
+      group.sort((a, b) => new Date(a.lastSubmissionAt).getTime() - new Date(b.lastSubmissionAt).getTime()),
```
rohitpaulk, line 19: Is `a.lastSubmissionAt` guaranteed to be not null?

rohitpaulk: Okay looks like all the records we have at the moment have this as non-null. Let's add a database migration in core so that this is codified (and we never create invalid records in the future)

app/models/course-participation.ts
rohitpaulk: Ignore, this seems to be non-null for all records. Noted this in another comment: 

> Okay looks like all the records we have at the moment have this as non-null. Let's add a database migration in core so that this is codified (and we never create invalid records in the future)

app/components/user-page/course-progress-list.ts
```diff
@@ -0,0 +1,43 @@
...
+  Element: HTMLDivElement;
+
+  Args: {
+    user: UserModel;
+  };
+};
+
+export default class CourseProgressListComponent extends Component<Signature> {
+  get courseParticipationGroups() {
+    const groupedCourseParticipations: CourseParticipationModel[][] = Object.values(groupBy(this.args.user.courseParticipations, 'course'));
```
rohitpaulk, line 16:
```suggestion
    const participationsGroupedByCourse: CourseParticipationModel[][] = Object.values(groupBy(this.args.user.courseParticipations, 'course'));
```

More consistent, it's easier to figure out the difference between this variable and the one below if the only difference in their name is the "sorted" prefix.

rohitpaulk (approved):

tests/pages/user-page.js
```diff
@@ -11,5 +12,7 @@ export default createPage({
     hover: triggerable('mouseenter'),
   },
 
+  courseProgressListItems: collection('[data-test-course-name]', CourseProgressListItem),
```
ryan-gang, line 15: In https://github.com/codecrafters-io/frontend/pull/1386#discussion_r1504967545,
did you mean this @rohitpaulk ?

rohitpaulk: @ryan-gang yes, doesn't necessarily have to be a separate file, but that's the idea. Example: 

Screenshot 2024-02-28 at 11 03 30

The usage you have here seems wrong btw, the collection selector should be something like `[data-course-progress-list-item]`

PR #1403, opened 2024-03-04
CC-1018: Update screencast time format in preview page.

ryan-gang:
feat: Add support for displaying formatted duration of screencasts in minutes, hours, and seconds

**Checklist**:

## Summary by CodeRabbit

- **New Features**
	- Introduced a new Glimmer component for displaying course stage screencast previews with enhanced functionality.
	- Added a helper function to format screencast durations in a human-readable way.
- **Enhancements**
	- Improved the display of screencast durations to be more user-friendly.
- **Bug Fixes**
	- Removed unnecessary comments from the course tester version activation date display.
- **Tests**
	- Updated tests to include new parameter for screencast duration and added assertions for formatted duration text.
- **Documentation**
	- Enhanced documentation and type declarations for new and updated helper functions.


app/components/course-stage-screencast-preview.hbs
```diff
@@ -20,7 +20,7 @@
     <div class="flex gap-3 mt-3">
       <div class="flex items-center">
         {{svg-jar "clock" class="w-4 text-gray-300 mr-1.5"}}
-        <div class="text-gray-600 text-xs">{{round (div @screencast.durationInSeconds 60)}} minutes</div>
+        <div class="text-gray-600 text-xs" data-test-screencasts-formatted-duration>{{format-time-duration @screencast.durationInSeconds}}</div>
```
rohitpaulk, line 23: @ryan-gang let's convert this component to typescript & remove the `@glint-nocheck` (you can replace it with `{{@glint-expect-error}}` if any sub-components aren't type-safe. 

As a practice, let's always convert components when we make changes to them. 

In this case for example, the newly added helper needs to be added to the Glint registry - something that we didn't catch because this component doesn't have typechecking

rohitpaulk, line 23:
```suggestion
        <div class="text-gray-600 text-xs" data-test-screencasts-formatted-duration>{{format-duration @screencast.durationInSeconds}}</div>
```

Duration seems sufficient as a name - don't think we need the "time" prefix. 

For example, the underlying date-fns library has formatDuration, and then there's  https://api.rubyonrails.org/classes/ActiveSupport/Duration.html. Neither have the "time" prefix

app/helpers/format-time-duration.ts
```diff
@@ -0,0 +1,12 @@
+import { helper } from '@ember/component/helper';
+import { formatDuration, intervalToDuration } from 'date-fns';
+
+export default helper(function formatTimeDuration([timeDurationInSeconds]: [number] /*, named*/) {
+  const duration = intervalToDuration({ start: 0, end: timeDurationInSeconds * 1000 });
+
+  if (timeDurationInSeconds > 60) {
+    duration.seconds = 0;
```
rohitpaulk, line 8: This seems kind of specific to the current use-case... If we had another place where we wanted to show something like "time since last attempt", it's possible that we want a more accurate timestamp like "1m 34s"

Can we try to make the interface more generic? I'd look at other similar functions you can find in OSS repos to figure out what a good interface could be.

rohitpaulk: If naming is hard here, I'd instead recommend doing this logic on `@screencast`. You could add `@screencast.roundedDurationInSecondsForDisplay` or something like that - and have it return the actual value if < 1m, and the rounded value if > 1m.

rohitpaulk (changes requested):
Comments added

app/components/course-stage-screencast-preview.ts
```diff
@@ -0,0 +1,22 @@
...
+interface Signature {
+  Element: HTMLDivElement;
+
+  Args: {
+    screencast: CourseStageScreencastModel;
+  };
+}
+
+export default class CourseStageScreencastPlayer extends Component<Signature> {
+  get roundedDurationToNearestMinute() {
```
rohitpaulk, line 13: Grammatically this name doesn't make sense - it's a valid "sentence" but not a valid "object". Also it removes the "InSeconds", so it's isn't clear what unit the new value is in - is it minutes? is it seconds?

Some valid alternatives: 

- `durationInSecondsRoundedToNearestMinute`
- `durationInSecondsForDisplay`
- `roundedDurationInSeconds`

My thought process: The #1 aspect to convey using this name is that it is a number and unit is seconds. #2 is how it is different from `@screencast.durationInSeconds` - in this case, it's because it's rounded / modified in some way for display purposes.

app/helpers/format-duration.ts
```diff
@@ -0,0 +1,8 @@
+import { helper } from '@ember/component/helper';
+import { formatDuration as libFormatDuration, intervalToDuration } from 'date-fns';
+
+export default helper(function formatDuration([timeDurationInSeconds]: [number] /*, named*/) {
```
rohitpaulk, line 4:
```suggestion
export default helper(function formatDuration([durationInSeconds]: [number] /*, named*/) {
```

types/glint.d.ts
```diff
@@ -12,6 +12,7 @@ declare module '@glint/environment-ember-loose/registry' {
     capitalize: HelperLike<{ Return: string; Args: { Positional: [string] } }>;
     'date-format': HelperLike<{ Return: string; Args: { Positional: [Date]; Named: { format?: string } } }>;
     'date-from-now': HelperLike<{ Return: string; Args: { Positional: [Date]; Named: { currentDate?: Date } } }>;
+    'format-duration': HelperLike<{ Return: string; Args: { Positional: [number] } }>;
```
rohitpaulk, line 15: This file is used for helpers that we don't control (from libraries) - for helpers in our app, we place the glint registrations directly in the file like this: 

https://github.com/codecrafters-io/frontend/blob/fd9d5570c1082fd6b1a0825d3b290c2e826b5487/app/helpers/current-user-is-course-author.ts#L25-L29

By doing this, we don't have to specify the interface separately, it is inferred from the function. This prevents mistakes where the interface specified here doesn't match the actual helper.
 
Let's move this registration into the helper file, and let's also move `date-format` and `date-from-now` into files please - they were previously part of a library we were using and were converted to helpers later. All the other helpers in this file _should_ be from libraries.

rohitpaulk (approved):

PR #1424, opened 2024-03-07
CC-1054 display extension ideas in the configure course extensions modal

libmartinito:
## Summary by CodeRabbit

- **New Features**
	- Added a voting section for future course extensions, allowing users to submit and view extension ideas.
	- Implemented sorting for course extension ideas based on user votes.
- **Tests**
	- Enhanced tests to cover the creation and display of course extension ideas.

### DIFF
--- a/app/components/course-page/configure-extensions-modal.hbs
+++ b/app/components/course-page/configure-extensions-modal.hbs
@@ -7,9 +7,28 @@
     Extensions are sets of additional stages that can be added to your repository. You can add &amp; remove extensions at any time.
   </Alert>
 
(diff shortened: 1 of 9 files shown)


rohitpaulk:
@libmartinito let's filter out "released" extensions from the list here: 

Screenshot 2024-03-07 at 21 30 03

rohitpaulk (approved):

rohitpaulk:
I also think we need a test here, at least to make sure that the data loading etc. is correct - the components are already tested in the vote page so we don't need extensive tests for all cases. Just one end-to-end test that ensure the cards appear should do.

PR #1467, opened 2024-03-18
refactor: Updated CourseParticipationModel to include completedStageSlugs.

ryan-gang:
Removed currentStage property from CourseParticipationModel and added completedStageSlugs property to store an array of completed stage slugs. This change allows for a more accurate calculation of completed stages count based on the slugs.

**Checklist**:

### DIFF
--- a/app/components/user-page/course-progress-list-item.js
+++ b/app/components/user-page/course-progress-list-item.js
@@ -15,7 +15,15 @@ export default class CourseProgressListItemComponent extends Component {
   }
 
   get completedStagesCount() {
-    return Math.max(...this.args.courseParticipations.mapBy('currentStage.position')) - 1;
+    let stagesCount = 0;
+
+    this.args.courseParticipations.mapBy('completedStageSlugs').forEach((stage_list) => {
+      if (stage_list) {
+        stagesCount = Math.max(stagesCount, stage_list.split(',').uniq().length);
+      }
+    });
+
+    return stagesCount;
   }
 
   get course() {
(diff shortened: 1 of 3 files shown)


app/models/course-participation.ts
```diff
@@ -1,17 +1,16 @@
...
 
 export default class CourseParticipationModel extends Model {
   @belongsTo('course', { async: false, inverse: null }) declare course: CourseModel;
-  @belongsTo('course-stage', { async: false, inverse: null }) declare currentStage: CourseStageModel;
   @belongsTo('user', { async: false, inverse: 'courseParticipations' }) declare user: UserModel;
   @belongsTo('language', { async: false, inverse: null }) declare language: LanguageModel;
 
   @attr('date') declare completedAt: Date | null;
   @attr('date') declare lastSubmissionAt: Date;
+  @attr('string') declare completedStageSlugs: string;
```
rohitpaulk, line 13: Let's send an array from the API. Look at `LeaderboardEntry#relatedLanguageSlugs` for an example.

app/components/user-page/course-progress-list-item.js
```diff
@@ -15,7 +15,15 @@ export default class CourseProgressListItemComponent extends Component {
   }
 
   get completedStagesCount() {
```
rohitpaulk, line 17: Let's handle de-duplication in the API - this should just be `this.completedStageSlugs.length`

rohitpaulk (approved):

PR #1556, opened 2024-04-09
CC-1115 add toggles to extension headings in stage list

libmartinito:
## Summary by CodeRabbit

- **New Features**
	- Enhanced user control over course customization with the introduction of a toggle feature for extensions in the course sidebar.
- **Enhancements**
	- Improved user interaction by adjusting the toggle button size and adding dynamic tooltips for clarity.
- **Tests**
	- Expanded test coverage to ensure reliable enabling/disabling of course extensions.
- **Documentation**
	- Updated component documentation to reflect new properties and functionalities.

### DIFF
--- a/app/components/course-page/step-list.hbs
+++ b/app/components/course-page/step-list.hbs
@@ -42,9 +42,9 @@
           <Toggle
             @isDisabled={{get @activeRepository "isNew"}}
             @isOn={{(not @configureExtensionsModalIsOpen)}}
-            @size='small'
+            @size="small"
             @tooltipCopy={{this.toggleButtonTooltipCopy}}
-            {{on 'click' @onConfigureExtensionsButtonClick}}
+            {{on "click" @onConfigureExtensionsButtonClick}}
             data-test-configure-extensions-toggle
           />
         </h3>
--- a/app/components/course-page/step-list.hbs
+++ b/app/components/course-page/step-list.hbs
@@ -36,8 +36,10 @@
   {{#each this.otherStepGroups as |stepGroup|}}
     {{#if stepGroup.title}}
       <div class="px-3 py-2">
-        <h3 class="flex justify-between items-center text-xs uppercase font-semibold text-gray-400 border-b pb-0.5">
-          {{stepGroup.title}}
+        <div class="flex justify-between items-center text-gray-400 border-b pb-0.5">
+          <h3 class="text-xs uppercase font-semibold">
+            {{stepGroup.title}}
+          </h3>
 
           <div>
             <Toggle
@@ -50,7 +52,7 @@
 
             <EmberTooltip @text={{this.toggleButtonTooltipCopy}} />
           </div>
-        </h3>
+        </div>
       </div>
     {{/if}}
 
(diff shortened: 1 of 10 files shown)


app/components/course-page/step-list.hbs
```diff
@@ -36,7 +36,18 @@
   {{#each this.otherStepGroups as |stepGroup|}}
     {{#if stepGroup.title}}
       <div class="px-3 py-2">
-        <h3 class="text-xs uppercase font-semibold text-gray-400 border-b">{{stepGroup.title}}</h3>
+        <h3 class="flex justify-between items-center text-xs uppercase font-semibold text-gray-400 border-b pb-0.5">
+          {{stepGroup.title}}
+
+          <Toggle
+            @isDisabled={{get @activeRepository "isNew"}}
+            @isOn={{(not @configureExtensionsModalIsOpen)}}
```
rohitpaulk, line 44: @libmartinito the toggle state here is supposed to indicate whether the extension is enabled, so this can just be `true`, no?

app/components/toggle.hbs
```diff
@@ -35,4 +40,8 @@
       </svg>
     </span>
   </span>
+
+  {{#if (and @tooltipCopy (not-eq @tooltipCopy ""))}}
+    <EmberTooltip @tooltipClass="normal-case font-normal" @text={{@tooltipCopy}} />
```
rohitpaulk, line 45: @libmartinito this isn't a very "flexible" interface - every time we have to pass an option to the `EmberTooltip` component (you might want to configure the `@position` for example), you have to add it to the `Toggle` component's interface. 

The standard way to do this is to instead wrap this in the caller. Example: 

```hbs
<div class="flex">
  <Toggle />
  <EmberTooltip />
</div>
```

This gives the caller full flexibility on what they pass to `EmberTooltip` - it also gives them flexibility on whether they want to show a tooltip, popover etc.

app/components/toggle.ts
```diff
@@ -6,13 +6,28 @@ interface Signature {
   Args: {
     isDisabled?: boolean;
     isOn: boolean;
+    isVisible?: boolean;
```
rohitpaulk, line 9: Where is `isVisible` used?

app/components/toggle.hbs
rohitpaulk, line 45: @libmartinito why do we have to update `@tooltipClass` here? I don't remember doing this elsewhere

rohitpaulk: If this is to avoid outer classes messing with styling, then the correct fix is to instead change outer classes to not override styling (we can't possible account for the 100s of different text styles outer classes might apply - what if they override `color` tomorrow? or `text-decoration`?).

rohitpaulk (changes requested):

app/components/toggle.ts
rohitpaulk: In general it's a bad idea to accept `isVisible` for any component - a caller should just not render the component, or it should use styling to override (`class="hidden"` or `class="opacity-0"`)

rohitpaulk (approved):

PR #1616, opened 2024-04-22
CC-1149 expose pricing page to unauthenticated users

libmartinito:
## Summary by CodeRabbit

- **New Features**
    - Enhanced the payment page to display specific elements and tooltips for unauthenticated users.
    - Payment page now conditionally fetches course data and regional discounts based on user authentication status.
    - Unauthenticated users can now access the payment page.

- **Bug Fixes**
    - Improved handling of user authentication checks across the payment process.

- **Tests**
    - Updated acceptance tests to reflect new user access capabilities on the payment page.

### DIFF
--- a/app/controllers/pay.ts
+++ b/app/controllers/pay.ts
@@ -7,7 +7,6 @@ import type AuthenticatorService from 'codecrafters-frontend/services/authentica
 import type MonthlyChallengeBannerService from 'codecrafters-frontend/services/monthly-challenge-banner';
 import type RouterService from '@ember/routing/router-service';
 import type { ModelType } from 'codecrafters-frontend/routes/pay';
-import type UserModel from 'codecrafters-frontend/models/user';
 
 export default class PayController extends Controller {
   declare model: ModelType;
@@ -42,7 +41,7 @@ export default class PayController extends Controller {
   }
 
   get user() {
-    return this.authenticator.currentUser as UserModel | null;
+    return this.authenticator.currentUser;
   }
 
   @action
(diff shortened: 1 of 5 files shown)


app/controllers/pay.ts
```diff
@@ -26,29 +26,33 @@ export default class PayController extends Controller {
...
+    } else if (this.user?.isEligibleForEarlyBirdDiscount) {
       return 216;
     } else {
       return null;
     }
   }
 
   get user() {
-    return this.authenticator.currentUser as UserModel; // pay route is protected by auth, so currentUser is guaranteed to be a UserModel
+    return this.authenticator.currentUser as UserModel | null;
```
rohitpaulk, line 45:
```suggestion
    return this.authenticator.currentUser;
```

(avoid `as` ensures that if the contract changes in the future, we'll start seeing type errors)

app/routes/pay.ts
```diff
@@ -20,15 +22,20 @@ export default class PayRoute extends BaseRoute {
...
 
   async model() {
-    await this.authenticator.authenticate();
+    if (this.authenticator.isAuthenticated) {
+      return {
+        courses: await this.store.findAll('course'), // For testimonials
+        regionalDiscount: await this.store.createRecord('regional-discount').fetchCurrent(),
+      };
+    } else {
+      if (this.authenticator.currentUser && this.authenticator.currentUser.hasActiveSubscription) {
```
rohitpaulk, line 31: This is in the `else` clause for `authenticator.isAuthenticated`, doesn't look like this condition will ever trigger then? 

Let's add a test for this - a paying user viewing `/pay` should be redirected.

rohitpaulk (changes requested):
Needs two additional tests: 

- User is redirected if they have a membership (I think this is broken at the moment)
- User is sent to login if they aren't logged in

PR #1664, opened 2024-04-28
CC-1142 update dates in real time

libmartinito:
## Summary by CodeRabbit

- **New Features**
  - Introduced `TimeService` to manage and update the current time at regular intervals.
  - Refactored date calculation to use a class-based helper `DateFromNow` for more accurate time distance measurement.

- **Bug Fixes**
  - Improved time-related functionality in tests with `FakeDateService` for consistent and reliable testing.

- **Chores**
  - Updated dependencies to include `@sinonjs/fake-timers` and `@types/sinonjs__fake-timers` for enhanced time manipulation during tests.


libmartinito:
@rohitpaulk could you please take a look at the approach before i go through the broken tests?

app/helpers/date-from-now.ts
```diff
@@ -1,21 +1,41 @@
...
-});
+  compute(positional: [Date]) {
+    let [date] = positional;
+    this.scheduleRecompute();
+    return formatDistanceStrictWithOptions({ addSuffix: true }, new Date(), date);
+  }
+
+  scheduleRecompute() {
+    this.clearTimer();
+    this.timer = window.setTimeout(() => this.recompute(), 1000);
```
rohitpaulk, line 28: @libmartinito one issue here is that different helpers will end up updating at different times - can we try to keep only one "timer" loop across all helpers? 

From linear: 

> Let's update this once every second, and let's make sure all timestamps across the app update at the same point of time so it isn't jarring.

app/services/time.ts
```diff
@@ -0,0 +1,17 @@
...
+  constructor() {
+    super();
+    this.updateTime();
+  }
+
+  updateTime() {
+    setInterval(() => {
+      this.currentTime = new Date();
+    }, 1000);
+  }
```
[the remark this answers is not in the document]
rohitpaulk: @libmartinito this is valid - in tests, services are instantiated and destroyed for each test, so you do need to account for cleanup

app/helpers/date-from-now.ts
```diff
@@ -1,21 +1,25 @@
...
+    Positional: Array<Date | null | undefined>;
   }
+}
 
-  return formatDistanceStrictWithOptions({ addSuffix: true }, currentDate, date);
-});
+export default class DateFromNow extends Helper<DateFromNowSignature> {
+  @service declare time: TimeService;
+
+  compute(positional: [Date]) {
```
rohitpaulk, line 15:
```suggestion
  compute(positional: DateFromNowSignature['Args']['Positional']) {
```

Note that we need to handle null / undefined values here, don't think that's done at the moment

rohitpaulk (approved):

PR #1672, opened 2024-04-30
CC-1194 allow admins to pin code examples

libmartinito:
## Summary by CodeRabbit

- **New Features**
	- Added a toggle button in the course admin section to pin code examples as editor's choice.
	- Introduced new API routes for fetching and updating course stage solutions and for solution comparisons.
	- Replaced the function in the `community-solution-card` component for publishing to GitHub.

- **Bug Fixes**
	- Ensured that pinning functionality correctly toggles and saves the pin status of code examples.

- **Tests**
	- Added acceptance tests for the new pinning functionality in the course admin section.
	- Created page object definitions for the course admin code example page to facilitate testing.

### DIFF
--- a/app/components/course-page/course-stage-step/community-solution-card/content.ts
+++ b/app/components/course-page/course-stage-step/community-solution-card/content.ts
@@ -17,7 +17,7 @@ type Signature = {
   Args: {
     solution: CommunityCourseStageSolutionModel;
     fileComparisons: FileComparison[];
-    onPublishToGithubButtonClick?: (event: MouseEvent) => void;
+    onPublishToGithubButtonClick?: () => void;
   };
 };
 
@@ -101,9 +101,9 @@ export default class CommunitySolutionCardContentComponent extends Component<Sig
   }
 
   @action
-  handlePublishToGithubButtonClick(event: MouseEvent) {
+  handlePublishToGithubButtonClick() {
     if (typeof this.args.onPublishToGithubButtonClick === 'function') {
-      this.args.onPublishToGithubButtonClick(event);
+      this.args.onPublishToGithubButtonClick();
     }
   }
 }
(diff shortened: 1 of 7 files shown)


app/components/course-page/course-stage-step/community-solution-card/content.ts
```diff
@@ -99,6 +99,13 @@ export default class CommunitySolutionCardContentComponent extends Component<Sig
       this.expandedUnchangedFilePaths = [...this.expandedUnchangedFilePaths, filePath];
     }
   }
+
+  @action
+  handlePublishToGithubButtonClick() {
+    if (typeof this.args.onPublishToGithubButtonClick === 'function') {
+      this.args.onPublishToGithubButtonClick();
+    }
```
libmartinito, line 107: not sure if this is the best way to handle this, but this was what i got to work. the on modifier errors out if we pass in a callback function that's potentially undefined

app/components/course-page/course-stage-step/community-solution-card/content.ts
```diff
@@ -99,6 +99,13 @@ export default class CommunitySolutionCardContentComponent extends Component<Sig
       this.expandedUnchangedFilePaths = [...this.expandedUnchangedFilePaths, filePath];
     }
   }
+
+  @action
+  handlePublishToGithubButtonClick() {
+    if (typeof this.args.onPublishToGithubButtonClick === 'function') {
```
rohitpaulk, line 105:
```suggestion
    if (this.args.onPublishToGithubButtonClick) {
```

(Can be simpler, we don't need to exactly check the type - it's going to be null/undefined if not passed in)

app/components/course-page/course-stage-step/community-solution-card/content.ts
rohitpaulk: This works! Added a comment on how it could be simpler

PR #1680, opened 2024-05-02
CC-1198 fix alignment of question options

libmartinito:
## Summary by CodeRabbit

- **Style**
	- Improved text alignment in question cards for better readability.

### DIFF
--- a/app/components/concept-page/question-card-option.hbs
+++ b/app/components/concept-page/question-card-option.hbs
@@ -1,7 +1,7 @@
 <button
   type="button"
   tabindex={{if @isSubmitted "-1" "0"}}
-  class="flex items-start gap-2 py-2
+  class="flex items-start text-left gap-2 py-2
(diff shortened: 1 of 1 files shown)


app/components/concept-page/question-card-option.hbs
```diff
@@ -23,7 +23,7 @@
       <div class="rounded-full w-3 h-3 bg-gray-500"></div>
     {{/if}}
   </div>
-  <div class="prose {{if (and this.isSubmitted this.isCorrect) 'text-gray-800 font-semibold' 'text-gray-800'}}">
+  <div class="prose text-left {{if (and this.isSubmitted this.isCorrect) 'text-gray-800 font-semibold' 'text-gray-800'}}">
```
rohitpaulk, line 26: @libmartinito can we add this to `button` instead? I think that's what causes the override in the first place (a `prose` _should_ be text-left by default?)

rohitpaulk (approved):

PR #1699, opened 2024-05-08
concept-question-keyboard-shortcuts

libmartinito:


app/components/concept-page/question-card.hbs
```diff
@@ -4,14 +4,19 @@
       {{markdown-to-html @question.queryMarkdown}}
     </div>
 
-    <div class="flex flex-col -mx-3 sm:-mx-4 gap-px" {{did-insert this.handleDidInsertOptionsList}}>
+    {{! template-lint-disable no-invalid-interactive }}
+    <div
+      class="question-card-option-container flex flex-col -mx-3 sm:-mx-4 gap-px"
+      {{did-insert this.handleDidInsertOptionsList}}
+      {{on "keydown" this.handleKeydown}}
```
rohitpaulk, line 11: @libmartinito can't this be implemented using `{{on-key}}`? (Look at how the other keyboard shortcuts like Enter / Backspace are implemented)

rohitpaulk (dismissed):
Note re: `{{on-key` added

libmartinito:
had to remove tests for navigation. problems were with how ember-cli-page-object behaves:
- for j/k: `triggerable` + keydown event doesn't accept lowercase keys
- for Enter: `triggerable` + keydown requires a selector where the element would be triggered. this is a problem because we removed explicit handling for keydown events with 'Enter' as this is an expected browser behavior

rohitpaulk:
@libmartinito Managed to add tests here, recorded videos:

- https://www.loom.com/share/13ce0573c4cd4dda949524488b3d835f
- https://www.loom.com/share/55f50d498fdd41d48d26352ef38f4cf4

app/components/concept-page/question-card.hbs
```diff
@@ -4,14 +4,27 @@
...
+      {{did-insert this.handleDidInsertOptionsList}}
+      {{! @glint-expect-error on-key modifier types aren't right? }}
+      {{on-key "k" this.handleMoveUp}}
+      {{! @glint-expect-error on-key modifier types aren't right? }}
+      {{on-key "j" this.handleMoveDown}}
+      {{! @glint-expect-error on-key modifier types aren't right? }}
+      {{on-key "ArrowUp" this.handleMoveUp}}
+      {{! @glint-expect-error on-key modifier types aren't right? }}
+      {{on-key "ArrowDown" this.handleMoveDown}}
+      {{on "keydown" this.handleKeydown}}
```
rohitpaulk, line 19: @libmartinito looks like we're still implementing keydown here? can't this be done using `on-key` too? Note that `on-key` doesn't need to be used as a modifier, it can be used as a regular helper too. 

So you can do something like: 

```
{{#each this.optionShortcuts as |shortcut|}}
  {{on-key shortcut.key (fn this.handleOptionShortcut(shortcut.optionIndex)}}
{{/each}}
```

(Naming could be better there, just wanted to illustrate how you can add dynamic shortcuts)
{{/each}}

rohitpaulk (changes requested):
We're still using `keydown` which is likely to be error prone and cause other issues - added note re: using `on-key` for the letters / numbers too

app/components/concept-page/question-card.ts
```diff
@@ -55,7 +63,37 @@ export default class QuestionCardComponent extends Component<Signature> {
   }
 
   @action
-  handleOptionClick(optionIndex: number) {
+  handleMoveDown() {
+    const currentFocusedOption = document.activeElement as HTMLElement;
```
rohitpaulk, line 67:
```suggestion
    const currentFocusedOption = document.activeElement;
```

It's wrong to assume that there's always a currently focused option - this will not be true.

rohitpaulk: (CodeRabbit highlighted this below)

app/components/concept-page/question-card.ts
```diff
@@ -55,7 +63,37 @@
...
+    const options = Array.from(document.querySelectorAll('[data-test-question-card-option]')) as HTMLElement[];
+    const currentFocusedOptionIndex = options.indexOf(currentFocusedOption);
+
+    if (currentFocusedOptionIndex < options.length - 1) {
+      options[currentFocusedOptionIndex + 1]!.focus();
+    } else {
+      options[0]!.focus();
+    }
+  }
```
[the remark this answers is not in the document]
rohitpaulk: @libmartinito ^ this is valid - the way we were using `as` was incorrect

rohitpaulk:
@libmartinito when using arrow keys, the page moves too - is this intended? I'd expect scrolling to be hijacked when a question card is "active" (i.e. after it's answered, arrow keys should work like normal but when it's active arrow keys should just select options and not move the page)

rohitpaulk (changes requested):
Added note re: arrow key behaviour

app/components/concept-page/question-card.hbs
```diff
@@ -4,17 +4,37 @@
       {{markdown-to-html @question.queryMarkdown}}
     </div>
 
-    <div class="flex flex-col -mx-3 sm:-mx-4 gap-px" {{did-insert this.handleDidInsertOptionsList}}>
+    {{! template-lint-disable no-invalid-interactive }}
+    <div
+      class="question-card-option-container flex flex-col -mx-3 sm:-mx-4 gap-px"
+      {{did-insert this.handleDidInsertOptionsList}}
+      {{! @glint-expect-error on-key modifier types aren't right? }}
```
rohitpaulk, line 11: We can avoid all of these warnings by using `{{on-key}}` as a helper instead, I don't see a particular reason to attach it to this container element

rohitpaulk (approved):

PR #1709, opened 2024-05-10
CC-1214 use completed stage slugs for leaderboard entry position handling

libmartinito:
## Summary by CodeRabbit

- **New Features**
	- Added a new property `completedStageSlugs` to track completed stages in course leaderboards.
	- Enhanced leaderboard entries to reflect the count of completed stages more accurately.

- **Bug Fixes**
	- Fixed inconsistencies in leaderboard data handling and display.

- **Refactor**
	- Simplified the logic for calculating completed stages count in various components.

- **Tests**
	- Updated and added test scenarios to validate the correct tracking and display of completed stages on leaderboards.

### DIFF
--- a/app/components/course-leaderboard.ts
+++ b/app/components/course-leaderboard.ts
@@ -95,6 +95,7 @@ export default class CourseLeaderboardComponent extends Component<Signature> {
       // TODO: Use "completed stages count" instead?
       return new CourseLeaderboardEntry({
         status: repository.lastSubmissionIsEvaluating ? 'evaluating' : repository.allStagesAreComplete ? 'completed' : 'idle',
+        completedStageSlugs: repository.completedStageSlugs,
         currentCourseStage: repository.currentStage || repository.course.firstStage,
         language: repository.language,
         user: repository.user,
@@ -127,6 +128,7 @@ export default class CourseLeaderboardComponent extends Component<Signature> {
       result.push(
         new CourseLeaderboardEntry({
           status: entriesForUser.isAny('status', 'evaluating') ? 'evaluating' : entryWithHighestCourseStage!.status,
+          completedStageSlugs: entryWithHighestCourseStage!.completedStageSlugs,
           currentCourseStage: entryWithHighestCourseStage!.currentCourseStage,
           language: entryWithHighestCourseStage!.language,
           user: entryWithHighestCourseStage!.user,
--- a/app/components/course-leaderboard.ts
+++ b/app/components/course-leaderboard.ts
@@ -92,7 +92,6 @@ export default class CourseLeaderboardComponent extends Component<Signature> {
     const allRepositories = this.args.repositories.toArray().concat([this.args.activeRepository]).uniq();
 
     return allRepositories.map((repository) => {
-      // TODO: Use "completed stages count" instead?
       return new CourseLeaderboardEntry({
         status: repository.lastSubmissionIsEvaluating ? 'evaluating' : repository.allStagesAreComplete ? 'completed' : 'idle',
         completedStageSlugs: repository.completedStageSlugs,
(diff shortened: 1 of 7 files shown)


app/models/course-leaderboard-entry.ts
```diff
@@ -14,11 +15,7 @@ export default class CourseLeaderboardEntry extends Model {
   @belongsTo('user', { async: false, inverse: null }) declare user: UserModel;
 
   get completedStagesCount() {
-    if (this.statusIsCompleted) {
-      return this.currentCourseStage.position;
-    } else {
-      return this.currentCourseStage.position - 1;
-    }
+    return this.completedStageSlugs?.length;
```
rohitpaulk, line 18:
```suggestion
    return this.completedStageSlugs.length;
```

(Given the type is `string[]`, we shouldn't need this)

app/components/course-leaderboard.ts
```diff
@@ -127,6 +128,7 @@ export default class CourseLeaderboardComponent extends Component<Signature> {
       result.push(
         new CourseLeaderboardEntry({
           status: entriesForUser.isAny('status', 'evaluating') ? 'evaluating' : entryWithHighestCourseStage!.status,
+          completedStageSlugs: entryWithHighestCourseStage!.completedStageSlugs,
```
rohitpaulk, line 131:
```suggestion
          completedStageSlugs: entryWithMostStageCompletions!.completedStageSlugs,
```

("HighestCourseStage" doesn't make sense anymore in this context. The logic is already accurate since `entryWithHighestCourseStage` is computed using `completedStagesCount`)

mirage/factories/course-stage-completion.js
```diff
@@ -19,6 +19,11 @@ export default Factory.extend({
       status: nextStage ? 'idle' : 'completed',
     });
 
+    courseStageCompletion.repository.update(
+      'completedStageSlugs',
```
rohitpaulk, line 23: @libmartinito isn't `repository.completedStageSlugs` a computed property? Not sure how updating that here helps - this seems inaccurate

mirage/factories/submission.js
```diff
@@ -85,6 +85,15 @@ export default Factory.extend({
           completedAt: submission.createdAt,
         });
       }
+
+      server.create('course-leaderboard-entry', {
+        status: submission.repository.lastSubmissionIsEvaluating ? 'evaluating' : submission.repository.allStagesAreComplete ? 'completed' : 'idle',
```
rohitpaulk, line 90: @libmartinito I have no idea how this works - `submission.repository` isn't an Ember data, model, it is a mirage model (it doesn't have all the computed properties like `allStagesAreComplete`, `lastSubmissionIsEvaluating` etc.). Can we double-check this?

app/components/course-leaderboard.ts
```diff
@@ -95,6 +95,7 @@ export default class CourseLeaderboardComponent extends Component<Signature> {
       // TODO: Use "completed stages count" instead?
```
rohitpaulk, line 95: We can remove this now

rohitpaulk (changes requested):
Added questions re: the test setup. I think the non-test code implementation here is accurate, the test setup I'm not so sure about. Requesting changes instead of approving, mostly for myself, I want to understand how this works

mirage/factories/submission.js
libmartinito: yep computed properties doesn't exist. the things you pointed out here (undefined check, updating a computed property, trying to create a leaderboard entry) originated from me thinking that course leaderboard entries aren't created. turns out, they're created in the course stage completions factory.

the problem that arose after creating a leaderboard entry here was i got one where the completedCourseStageSlugs were undefined. that's why i tried forcibly updating the value + adding an undefined check

rohitpaulk (approved):

PR #1774, opened 2024-05-26
Add a foundation mechanism for supporting Dark Mode

### LINKED ISSUE #1415: Dark Mode - foundations
We've got a couple of pages / areas within pages that are currently in dark mode. 

Examples: 

- https://app.codecrafters.io/badges
- https://app.codecrafters.io/contests/weekly-7

We want to add dark mode throughout the app, but will do so incrementally since there'll be multiple components that need to be edited to make things look good. 

As part of this task, let's add the "foundation" for this.

- We need dark mode state to be stored in local storage. The options are "light", "dark" and "system". The default will be "light". 
- For staff, we need a way to toggle this in the UI. Let's add another link to account dropdown [here](https://github.com/codecrafters-io/frontend/blob/fd9d5570c1082fd6b1a0825d3b290c2e826b5487/app/components/header/account-dropdown.hbs#L81), below the Test Sentry link.
- When this is toggled, we need the "dark" class to be applied / removed from the application layout. 
- Cleanup the `is-dark-mode` modifier - currently it's pretty hacky, uses intersectionobserver and what not
- Cleanup the `colorScheme` service, it currently uses hardcoded rules based on routes

Feel free to change the whole structure as much as you like, no need to stick to what we have already - most of this was added in a hurry and not properly thought out.  

Note: We need all of this to work in such a way that customers aren't affected, but staff can toggle this at any time and inspect how things look. At the same time, once a staff switches things back to "light" mode, this should be treated the same as a regular user's browser. It's okay for `badges`, `contests` etc. to always be in dark mode, we aren't going to support light mode for those endpoints anytime soon.

VasylMarchuk:
Closes #1415

### Brief

This adds a basic mechanism for supporting the Dark Mode, based on a preference read from local storage, combined with the current system preference.

### Details

- Added new `DarkModeService` for tracking Dark Mode environment settings
  - Used by application template to automatically add `.dark` class to the application's root wrapper element
  - Exposes only a few public actions/properties:
    - `get isEnabled` — whether dark mode should currently be used
    - `@action updateLocalStoragePreference` — write new preference to `localStorage` & trigger `isEnabled` recalculation
    - `@tracked localStoragePreference` — not supposed to be public, but can be used to see what `localStorage` preference is currently loaded
    - `@tracked systemPreference` — not supposed to be public, but can be used to see what system preference is currently loaded
- Added `RouteInfoMetadata` utility class for defining Dark Mode preferences in routes, currently with one property, `colorScheme`:
  - `RouteColorScheme.Light` (default) — indicates that the route & it's nested children only support Light Mode
  - `RouteColorScheme.Dark` — indicates that the route & it's nested children, only support Dark Mode
  - `RouteColorScheme.Both` — indicates that the route & it's nested children support both Dark and Light mode — local storage & system preferences start being used for them
  - Nesting is fully supported: `Light` route trees can be nested inside `Dark` & `Both`, `Dark` inside `Light` & `Both`, `Both` inside `Light` & `Dark`
  - Routes can use [`buildRouteInfoMetadata`](https://api.emberjs.com/ember/5.8/classes/Route/methods/buildRouteInfoMetadata?anchor=buildRouteInfoMetadata) hook to define the metadata
- Deleted `is-dark-mode` modifier
- Refactored code highlighting logic in `SyntaxHighlightedDiff` component
- Added support for multiple trigger classes in Tailwind's Dark Mode
  - currently using only `.dark` class, but possible to enable other options
- Added support for native Tailwind Dark Mode to all layout-related components
- Added route info metadata to Badges & Contests, returning `RouteColorScheme.Dark`
- Added an extremely simple component `DarkModeToggle` for changing the local storage preference
  - Added a new route `/demo/dark-mode-toggle` for testing the component
- Deleted the `ColorScheme` service, which was no longer used anywhere
- Added full support for Dark Mode to the Demo Route, returning `RouteColorScheme.Both`
- Found one usage of `SyntaxHighlightedDiff` in the manual “dark area” and fixed it, this required:
  - adding `forceDarkTheme` argument to `SyntaxHighlightedDiff` & `FileDiffCard` components
  - passing `true` for it in the `AutofixResult` component
- Added proper dark-mode-background classes to `SyntaxHighlightedDiff` component
- Global `glint.d.ts` registry now also inherits from `EmberConcurrencyRegistry`

### Demo

https://github.com/codecrafters-io/frontend/assets/493875/de94410c-1dea-476c-91f5-1eeab1b3cd73

https://github.com/codecrafters-io/frontend/assets/493875/2aeb4726-1b3b-47d9-84b3-472c2a858491

### Checklist

## Summary by CodeRabbit

- **New Features**
  - Introduced a dark mode toggle feature, allowing users to switch between "System," "Dark," "Light," and "Unset" preferences.
  - Added a new `DarkModeService` to manage dark mode preferences based on system settings and user choices.

- **Enhancements**
  - Updated various components to support and respect the dark mode settings.
  - Improved styling consistency across components for better readability in dark mode.

- **Bug Fixes**
  - Fixed class and color scheme handling in multiple components to ensure proper dark mode application.

- **Refactor**
  - Removed `isDark` property from several interfaces for cleaner code and better handling of dark mode via service.


app/components/syntax-highlighted-diff.ts
```diff
@@ -36,12 +64,6 @@ export default class SyntaxHighlightedDiffComponent extends Component<Signature>
   static LINES_AROUND_CHANGED_CHUNK = 3;
   static MIN_LINES_BETWEEN_CHUNKS_BEFORE_COLLAPSING = 4;
 
-  constructor(owner: unknown, args: Signature['Args']) {
-    super(owner, args);
-
-    this.highlightCode.perform();
-  }
-
```
VasylMarchuk, line 44: This is unnecessary, as invoking `highlightCode` at this point results into the task exiting immediately, because `#isDarkMode` is still `undefined`

app/components/syntax-highlighted-diff.ts
```diff
@@ -138,13 +160,8 @@ export default class SyntaxHighlightedDiffComponent extends Component<Signature>
 
   @action
   handleIsDarkModeUpdate(isDarkMode: boolean) {
-    if (isDarkMode !== this.isDarkMode) {
-      // Avoid re-use in same computation bug
-      next(() => {
-        this.isDarkMode = isDarkMode;
-        this.highlightCode.perform();
-      });
```
VasylMarchuk, line 146: Using `next` is unnecessary and harmful: it _leads_ to computation bugs and future errors in tests.

app/services/dark-mode.ts
```diff
@@ -0,0 +1,135 @@
...
+    // Load the localStorage preference from localStorage service
+    this.localStoragePreference = (this.localStorage.getItem(LOCAL_STORAGE_KEY) as LocalStoragePreference | null) || undefined;
+
+    // Create a media query to load current system Dark Mode preference
+    const mediaQuery = window.matchMedia('(prefers-color-scheme: dark)');
+
+    // Create a listener for media query result changes
+    const mqChangeListener = (e: MediaQueryListEvent) => {
+      this.systemPreference = e.matches ? 'dark' : 'light';
+      this.invokeUpdateListeners();
```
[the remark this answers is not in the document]
VasylMarchuk: Good luck testing media queries! 😉

app/services/dark-mode.ts
```diff
@@ -0,0 +1,135 @@
...
+  }
+
+  /**
+   * Writes the new Dark Mode preference to localStorage and executes all currently
+   * registered update listeners, passing them a new value
+   * @param {'system'|'dark'|'light'|null} newValue New dark mode preference
+   */
+  @action updateLocalStoragePreference(newValue?: LocalStoragePreference) {
+    if (!newValue) {
+      this.localStorage.removeItem(LOCAL_STORAGE_KEY);
```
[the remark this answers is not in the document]
VasylMarchuk: I don't think we've settled on proper local storage isolation in the tests environment. And this will require writing the preference to real local storage during tests — which messes up things for `localhost:4200` (at least)

app/components/tertiary-link-button.hbs
```diff
@@ -5,9 +5,7 @@
   @size={{@size}}
   @query={{@query}}
   @shouldOpenInNewTab={{@shouldOpenInNewTab}}
-  class="{{if @isDark 'bg-transparent' 'bg-white hover:bg-gray-50'}}
-    {{if @isDark 'text-gray-200 hover:text-white' 'text-gray-900'}}
-    {{if @isDark 'border-gray-300 hover:border-gray-200' 'border-gray-300'}}
+  class="bg-white dark:bg-transparent text-gray-900 dark:text-gray-200 border-gray-300 dark:border-gray-300 light:hover:bg-gray-50 dark:hover:text-white dark:hover:border-gray-200
```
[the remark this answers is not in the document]
VasylMarchuk: `hover:bg-gray-50` was previously NOT set when in Dark Mode, therefore listed explicitly for `light:` mode here

app/modifiers/is-dark-mode.ts
```diff
@@ -1,56 +1,139 @@
...
-    instance.observer.disconnect();
-    instance.observer = undefined;
-  }
-}
-
+/**
+ * Invokes a callback on initial load and when Dark Mode environment changes,
+ * passing it `isDarkMode: boolean` as argument.
+ * Calculated based on element's DOM parents with `.dark` class.
+ * Relies on DarkModeService and/or DOM Mutation Obsersers for monitoring changes.
```
VasylMarchuk, line 53:
```suggestion
 * Relies on DarkModeService and/or DOM Mutation Observers for monitoring changes.
```

app/modifiers/is-dark-mode.ts
```diff
@@ -1,56 +1,139 @@
...
+  /**
+   * Callback to invoke on initial load and when Dark Mode environment changes
+   * @private
+   */
+  #callback?: (isDarkMode: boolean) => void;
 
-  eventHandlerTask = task({ keepLatest: true }, async () => {
-    const isDarkMode = !!this.element!.closest('.dark');
+  /**
+   * Elemnent which the modifier was used on
```
VasylMarchuk, line 69:
```suggestion
   * Element which the modifier was used on
```

app/services/dark-mode.ts
```diff
@@ -0,0 +1,135 @@
...
+  /**
+   * Registers a callback to invoke when Dark Mode environment changes
+   * @param listener method to invoke, new boolean value will be passed as first argument
+   */
+  @action registerUpdateListener(listener: UpdateListener) {
+    this.#registeredListeners.push(listener);
+  }
+
+  /**
+   * Unregisters a callback previsouly resigtered with `registerUpdateListener`
```
VasylMarchuk, line 105:
```suggestion
   * Unregisters a callback previsouly registered with `registerUpdateListener`
```

VasylMarchuk (commented):
Typos review

app/components/syntax-highlighted-diff.ts
```diff
@@ -27,7 +43,19 @@ export default class SyntaxHighlightedDiffComponent extends Component<Signature>
...
+   * and used by `highlightCode` task to decide which mode to format code with
+   * @private
+   */
+  #isDarkMode: boolean | undefined = undefined;
+
+  /**
+   * The last Dark Mode environment setting that was used to format code
+   * @private
+   */
+  #isDarkModeRendered: boolean | undefined = undefined;
```
rohitpaulk, line 58: In terms of purpose, this seems similar to `asyncHighlightedCode`. Maybe we can name this similar to that?

app/components/syntax-highlighted-diff.ts
```diff
@@ -160,28 +177,53 @@ export default class SyntaxHighlightedDiffComponent extends Component<Signature>
     }
   }
 
-  highlightCode = task({ keepLatest: true }, async (): Promise<void> => {
-    if (this.isDarkMode === undefined) {
-      return; // Not ready to highlight yet
+  highlightCode = task({ [HIGHLIGHT_CODE_TASK_TIMEOUT_METHOD]: true }, async (): Promise<void> => {
+    // For "restartable" to have immediate effect for sequential calls
+    if (HIGHLIGHT_CODE_TASK_TIMEOUT_METHOD === 'restartable') {
```
rohitpaulk, line 182: Ignoring this since we're most likely removing restartable

app/modifiers/is-dark-mode.ts
```diff
@@ -1,56 +1,139 @@
...
+ * Enable or disable subscribing to `DarkModeService.isEnabled` changes
+ */
+const DARK_MODE_UPDATE_LISTENER_ENABLED: boolean = true;
+
+/**
+ * Method to use for delaying `triggerCallback` task execution:
+ * - `restartable` cancels all queued calls before they have a chance to execute, executes only the latest call: causes much less unnecessary task executions
+ * - `keepLatest` always executes first & last queued calls, cancelling the ones in-between: might be more reliable but always triggers rendering twice
+ */
+const CALLBACK_TASK_TIMEOUT_METHOD: 'keepLatest' | 'restartable' = 'restartable';
```
rohitpaulk, line 22: @VasylMarchuk same as above - let's remove any configuration options like this and just choose one

app/modifiers/is-dark-mode.ts
```diff
@@ -1,56 +1,139 @@
...
 
+/**
+ * Enable or disable the DOM Mutation Observer
+ */
+const DOM_MUTATION_OBSERVER_ENABLED: boolean = false; // TODO: Figure out an efficient way to enable the observer while avoiding too many calls
+
+/**
+ * Enable or disable subscribing to `DarkModeService.isEnabled` changes
+ */
+const DARK_MODE_UPDATE_LISTENER_ENABLED: boolean = true;
```
rohitpaulk, line 15: Avoid configuration options as far as possible, I'd suggest just removing this and keep it enabled

app/modifiers/is-dark-mode.ts
```diff
@@ -1,56 +1,139 @@
...
+/**
+ * Timeout used for `keepLatest` method when `DOM_MUTATION_OBSERVER_ENABLED` is true
+ */
+const CALLBACK_TASK_TIMEOUT_POST: number = 100;
+
+/**
+ * Name of element's attribute for storing the currently reported state of Dark Mode,
+ * to avoid invoking callbacks again with same value and to prevent double-renders
+ */
+const CACHED_STATE_ATTRIBUTE_NAME = 'data-reported-dark-mode';
```
rohitpaulk, line 38: Does this have to be stored in the DOM? Can it be stored on the modifier class, something like `@tracked lastCallbackValue`?

app/modifiers/is-dark-mode.ts
```diff
@@ -1,56 +1,139 @@
...
-import { task } from 'ember-concurrency';
+import { service } from '@ember/service';
+import { task, timeout } from 'ember-concurrency';
 import Modifier from 'ember-modifier';
+import type DarkModeService from 'codecrafters-frontend/services/dark-mode';
 
+/**
+ * Enable or disable the DOM Mutation Observer
+ */
+const DOM_MUTATION_OBSERVER_ENABLED: boolean = false; // TODO: Figure out an efficient way to enable the observer while avoiding too many calls
```
rohitpaulk, line 10: Let's just comment out code instead - all these configuration options make it harder to read code IMO, too many cases to think about

app/routes/contests.ts
```diff
@@ -15,6 +16,10 @@ export default class ContestsRoute extends BaseRoute {
   @service declare router: RouterService;
   @service declare store: Store;
 
+  buildRouteInfoMetadata(): RouteInfoMetadata {
+    return new RouteInfoMetadata({ isDarkRoute: true });
```
rohitpaulk, line 20: Nice! Didn't know this was a thing.

app/routes/badges.js
```diff
@@ -1,12 +1,17 @@
...
 import BaseRoute from 'codecrafters-frontend/utils/base-route';
+import RouteInfoMetadata from 'codecrafters-frontend/utils/route-info-metadata';
 import RSVP from 'rsvp';
 
 export default class BadgesRoute extends BaseRoute {
   allowsAnonymousAccess = true;
   @service authenticator;
   @service store;
 
+  buildRouteInfoMetadata() {
```
rohitpaulk, line 11: Nice! Didn't know this was a thing.

app/services/dark-mode.ts
```diff
@@ -0,0 +1,135 @@
...
+   */
+  get isEnabled(): boolean {
+    return this.#isVisitingDarkRoute || this.#isLocalStorageEnabled;
+  }
+
+  /**
+   * Returns whether Dark Mode should be enabled based on localStorage & system preference
+   * @private
+   */
+  get #isLocalStorageEnabled(): boolean {
```
rohitpaulk, line 75: Name is misleading - could be something like `isEnabledViaPreferences`

app/templates/application.hbs
```diff
@@ -2,7 +2,7 @@
 
 {{page-title "CodeCrafters"}}
 
-<div class="bg-gray-50 min-h-screen">
+<div class="application-container bg-gray-50 min-h-screen {{if this.darkMode.isEnabled 'dark'}}">
```
rohitpaulk, line 5: Now that I look at this with fresh eyes, it feels a bit confusing that `darkMode.isEnabled` and the value from `{{is-dark-mode` can be different. The names don't convey that they check different things. Not sure of a better name though, one idea was to make the modifier `{{is-within-dark-area ...}}`.

rohitpaulk (changes requested):
Added one round of comments!

Primary items: 

- remove all configuration options unless absolutely necessary, just pick one and roll with it.
- ensure that we're handling updates to `code` / `language` etc. when running `highlightCode`

app/components/syntax-highlighted-diff.ts
```diff
@@ -160,28 +177,53 @@ export default class SyntaxHighlightedDiffComponent extends Component<Signature>
...
+      return;
+    }
+
+    // Return if nothing changed since last render
+    if (this.#isDarkMode === this.#isDarkModeRendered) {
+      return;
+    }
+
+    // Remember which mode we've last used for rendering
+    this.#isDarkModeRendered = this.#isDarkMode;
```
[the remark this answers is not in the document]
VasylMarchuk: Was thinking about it myself, will do! If we had error hanging then this would make sense, but we don't.

app/components/syntax-highlighted-diff.ts
```diff
@@ -8,8 +8,24 @@ import { escapeHtml, groupBy, zip } from 'codecrafters-frontend/utils/lodash-uti
...
 import { tracked } from '@glimmer/tracking';
 import { transformerNotationDiff } from '@shikijs/transformers';
-import { next } from '@ember/runloop';
-import { task } from 'ember-concurrency';
+import { task, timeout } from 'ember-concurrency';
+
+/**
+ * Method to use for delaying `highlightCode` task execution:
+ * - `restartable` cancels all queued calls before they have a chance to execute, executes only the latest call: causes much less unnecessary task executions
+ * - `keepLatest` always executes first & last queued calls, cancelling the ones in-between: might be more reliable but always triggers rendering twice
```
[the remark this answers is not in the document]
VasylMarchuk: ##Was just experimenting with both, I've got a branch with a lot of debug logging added, and wanted to record this in git history for later, maybe

app/components/syntax-highlighted-diff.ts
```diff
@@ -27,7 +43,19 @@ export default class SyntaxHighlightedDiffComponent extends Component<Signature>
...
   @tracked asyncHighlightedCode: string | null = null;
   @tracked lineNumberWithExpandedComments: number | null = null;
-  @tracked isDarkMode: boolean | undefined = undefined;
+
+  /**
+   * Current Dark Mode environment setting, reported by `is-dark-mode` modifier
+   * and used by `highlightCode` task to decide which mode to format code with
+   * @private
+   */
+  #isDarkMode: boolean | undefined = undefined;
```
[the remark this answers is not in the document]
VasylMarchuk: Thought about it too, but the task is sometimes called without any args (when `code` changes, for example), and thus we need to remember both. Another option would be to pass the argument everywhere.

app/components/syntax-highlighted-diff.ts
```diff
@@ -160,28 +177,53 @@ export default class SyntaxHighlightedDiffComponent extends Component<Signature>
...
+      await timeout(HIGHLIGHT_CODE_TASK_TIMEOUT_PRE);
     }
 
+    // Return if not ready to highlight yet
+    if (this.#isDarkMode === undefined) {
+      return;
+    }
+
+    // Return if nothing changed since last render
+    if (this.#isDarkMode === this.#isDarkModeRendered) {
```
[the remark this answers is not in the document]
VasylMarchuk: We already have an action `handleDidUpdateCode` which will re-trigger the `formatCode` task

app/modifiers/is-dark-mode.ts
VasylMarchuk: Will try!

app/routes/contests.ts
VasylMarchuk: Me neither, it was a draft RFC for a long time, then finally got accepted & merged at some point

app/templates/application.hbs
VasylMarchuk: `did-load-dark-parents` would express exactly what it does, and it's a modifier, so the name has to be a verb...

VasylMarchuk: In general, I would ditch dark areas and just hardcode dark theme in them where needed, somehow.


Having to scan the DOM parents for `.dark` class makes things 9 levels more complicated: this is the only reason this modifier is needed. Otherwise the logic everywhere could have been as simple as:

```
<CodeMirror @theme={{this.darkMode.isEnabled 'githubDark' : 'githubLight'}}
```

No need for modifier + action to handle it's callback

app/components/syntax-highlighted-diff.ts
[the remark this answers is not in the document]
rohitpaulk: Aha, yeah hadn't thought of that. Okay looks like this'll change quite a bit now that we're using args instead - will wait to see what that looks like

app/templates/application.hbs
rohitpaulk: Discussed on slack - we're going forward with ditching the modifier and instead just hardcoding `@theme` for syntax highlighted diffs in dark areas

app/components/syntax-highlighted-diff.ts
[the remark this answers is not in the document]
VasylMarchuk: I reworked this into the following interface and class:

```ts
interface HighlightCodeParams {
  isDarkMode?: boolean;
  language?: string;
  codeWithoutDiffMarkers?: string;
}

class HighlightCodeParamsCache implements HighlightCodeParams {
  isDarkMode?: boolean;
  language?: string;
  codeWithoutDiffMarkers?: string;

  compareWithParams({ isDarkMode, language, codeWithoutDiffMarkers }: HighlightCodeParams): boolean {
    return this.isDarkMode === isDarkMode && this.language === language && this.codeWithoutDiffMarkers === codeWithoutDiffMarkers;
  }

  setParams({ isDarkMode, language, codeWithoutDiffMarkers }: HighlightCodeParams): void {
    this.isDarkMode = isDarkMode;
    this.language = language;
    this.codeWithoutDiffMarkers = codeWithoutDiffMarkers;
  }
}
```

and now all `highlightCode` task has to do is as simple/reliable/typescript-backed as:

```ts
// Return if nothing changed since last render
if (
  this.#lastHighlightCodeParams.compareWithParams({
    isDarkMode: isDarkMode,
    language: language,
    codeWithoutDiffMarkers: codeWithoutDiffMarkers,
  })
) {
  return;
}

// ...

// Remember all parameters used for this task run
this.#lastHighlightCodeParams.setParams({
  isDarkMode: isDarkMode,
  language: language,
  codeWithoutDiffMarkers: codeWithoutDiffMarkers,
});
```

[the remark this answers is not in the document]
VasylMarchuk: Also the original implementation was missing a `did-update` trigger for whenever `@language` argument changed, I added it.

app/components/syntax-highlighted-diff.ts
```diff
@@ -8,8 +8,24 @@ import { escapeHtml, groupBy, zip } from 'codecrafters-frontend/utils/lodash-uti
...
+
+/**
+ * Timeout used for `restartable` method
+ */
+const HIGHLIGHT_CODE_TASK_TIMEOUT_PRE: number = 10;
+
+/**
+ * Timeout used for `keepLatest` method
+ */
+const HIGHLIGHT_CODE_TASK_TIMEOUT_POST: number = 100;
```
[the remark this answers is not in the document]
VasylMarchuk: Renamed to `DELAY_BETWEEN_HIGHLIGHT_CODE_TASKS_IN_MS`

app/components/syntax-highlighted-diff.ts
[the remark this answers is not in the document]
VasylMarchuk: This is no longer needed, the component only relies on `DarkModeService.isEnabled`, which is `@tracked` and automatically triggers `this.highlightCode` via `did-update` mechanism. Removed.

app/components/syntax-highlighted-diff.ts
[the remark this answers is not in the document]
VasylMarchuk: Removed this property, and it's now part of `HighlightCodeParams` interface / `HighlightCodeParamsCache` class, stored in `SyntaxHighlightedDiffComponent.#lastHighlightCodeParams` property.

app/components/syntax-highlighted-diff.ts
VasylMarchuk: Renamed this to `#lastHighlightCodeParams`, and it's now an instance of `HighlightCodeParamsCache` class, which implements `HighlightCodeParams` interface.

app/modifiers/is-dark-mode.ts
VasylMarchuk: Removed this helper completely

app/components/syntax-highlighted-diff.ts
VasylMarchuk: Removed in favour of `keepLatest` method

app/modifiers/is-dark-mode.ts
VasylMarchuk: Removed this helper completely

app/modifiers/is-dark-mode.ts
VasylMarchuk: Removes this helper completely

app/services/dark-mode.ts
VasylMarchuk: Renamed to `#isEnabledViaPreferences`

app/components/syntax-highlighted-diff.ts
VasylMarchuk: Now using `did-insert` to call `hightlightCode` instead of the constructor

app/components/syntax-highlighted-diff.ts
```diff
@@ -161,27 +175,57 @@ export default class SyntaxHighlightedDiffComponent extends Component<Signature>
...
+
+    // Remember all parameters used for this task run
+    this.#lastHighlightCodeParams.setParams({
+      isDarkMode: isDarkMode,
+      language: language,
+      codeWithoutDiffMarkers: codeWithoutDiffMarkers,
+    });
+
+    // Ensure we don't run this task too often
+    await timeout(DELAY_BETWEEN_HIGHLIGHT_CODE_TASKS_IN_MS);
```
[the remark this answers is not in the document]
VasylMarchuk: The task code is complex because that's the nature of the task it handles. The code is well structured, variables are well named, comments are thoughtfully placed, and it takes 3 minutes to actually read and understand — nothing hard.

app/components/syntax-highlighted-diff.ts
```diff
@@ -23,11 +30,40 @@
...
+}
+
+class HighlightCodeParamsCache implements HighlightCodeParams {
+  isDarkMode?: boolean;
+  language?: string;
+  codeWithoutDiffMarkers?: string;
+
+  compareWithParams({ isDarkMode, language, codeWithoutDiffMarkers }: HighlightCodeParams): boolean {
+    return this.isDarkMode === isDarkMode && this.language === language && this.codeWithoutDiffMarkers === codeWithoutDiffMarkers;
+  }
```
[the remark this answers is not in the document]
VasylMarchuk: Good idea

rohitpaulk:
(dismissed review since we spoke about the naming change)

app/components/syntax-highlighted-diff.hbs
```diff
@@ -1,4 +1,12 @@
-<div class="overflow-x-scroll" {{did-update this.handleDidUpdateCode @code}} {{is-dark-mode this.handleIsDarkModeUpdate}} ...attributes>
+<div
+  class="overflow-x-scroll dark:text-gray-400 dark:bg-gray-900"
+  {{did-insert (perform this.highlightCode)}}
+  {{did-update (perform this.highlightCode) this.darkMode.isEnabled}}
```
rohitpaulk, line 4: I think that `did-update` takes multiple args, so we could just pass all of them in one go

app/components/syntax-highlighted-diff.ts
```diff
@@ -161,27 +184,62 @@ export default class SyntaxHighlightedDiffComponent extends Component<Signature>
...
+    // Return if nothing changed since last render
+    if (
+      this.#lastHighlightCodeParams.compareWithParams({
+        isDarkMode,
+        forceDarkTheme,
+        language,
+        codeWithoutDiffMarkers,
+      })
+    ) {
+      return;
```
rohitpaulk, line 203: What if `@code` has changed, but nothing else has? In that case don't we still need to run highlighting?

app/components/syntax-highlighted-diff.ts
```diff
@@ -195,7 +253,7 @@ export default class SyntaxHighlightedDiffComponent extends Component<Signature>
 }
 
 declare module '@glint/environment-ember-loose/registry' {
-  export default interface Registry {
+  export default interface Registry extends EmberConcurrencyRegistry {
```
rohitpaulk, line 256: Can we place this in `glint.d.ts` instead? Should be a global thing, not specific to `syntax-highlighted-diff`

app/services/dark-mode.ts
```diff
@@ -0,0 +1,138 @@
...
+  /**
+   * Currently loaded system preference value
+   */
+  @tracked systemPreference?: SystemPreference;
+
+  constructor(properties: object | undefined) {
+    super(properties);
+
+    // Load the localStorage preference from localStorage service
+    this.localStoragePreference = this.localStorage.getItem(LOCAL_STORAGE_KEY) as LocalStoragePreference;
```
rohitpaulk, line 33:
```suggestion
    this.localStoragePreference = this.localStorage.getItem(LOCAL_STORAGE_KEY) || undefined;
```

(Using `as xyz` could result in `localStoragePreference` being set to `null`, which doesn't agree w/ the types)

rohitpaulk (approved):

rohitpaulk:
Minor comments, looks good otherwise!

app/components/syntax-highlighted-diff.ts
VasylMarchuk: In this case `codeWithoutDiffMarkers` will change and trigger the update

app/components/syntax-highlighted-diff.hbs
VasylMarchuk: Oh, that's cool if it does now, certainly didn't before, will check!

app/services/dark-mode.ts
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    return false;
+  }
+
+  /**
+   * Returns whether Dark Mode should be enabled because the user is visiting a "dark route"
+   * @private
+   */
+  get #isVisitingDarkRoute(): boolean {
```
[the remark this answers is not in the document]
VasylMarchuk: Maybe `currentRouteRequiresDarkMode`? This will match well with `currentRouteSupportsDarkMode`, I think.

app/services/dark-mode.ts
VasylMarchuk: The type does allow `null` though:

image

app/components/syntax-highlighted-diff.ts
VasylMarchuk: I believe this is needed so that `perform` helper can be used in the template

VasylMarchuk: Will try to make this global

app/services/dark-mode.ts
rohitpaulk: Ah, ignore me - I thought I saw a `localStoragePreference?` somewhere. All good!

app/components/syntax-highlighted-diff.ts
rohitpaulk: Ah, didn't notice - all good!

app/services/dark-mode.ts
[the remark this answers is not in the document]
rohitpaulk: Yep, that works!

PR #1783, opened 2024-05-27
CC-1258 Fix bug with submitting course stage feedback on the last base stage

libmartinito:
**Checklist**:

## Summary by CodeRabbit

- **New Features**
  - Added the ability to submit course stage feedback after passing a base stage.

- **Bug Fixes**
  - Improved navigation within course stages for better accuracy and reliability.

### DIFF
--- a/app/controllers/course/stage/instructions.ts
+++ b/app/controllers/course/stage/instructions.ts
@@ -82,7 +82,7 @@ export default class CourseStageInstructionsController extends Controller {
 
     // If the "active step" is after a "BaseStagesCompletedStep", navigate to "BaseStagesCompletedStep" instead.
     if (nextStep?.type === 'BaseStagesCompletedStep' && stepAfterNextStep === activeStep) {
-      this.router.transitionTo(nextStep.routeParams.route, nextStep.routeParams.models);
+      this.router.transitionTo(nextStep.routeParams.route, nextStep.routeParams.models[0]!);
     }
   }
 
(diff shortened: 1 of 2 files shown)


app/controllers/course/stage/instructions.ts
```diff
@@ -82,7 +82,7 @@ export default class CourseStageInstructionsController extends Controller {
 
     // If the "active step" is after a "BaseStagesCompletedStep", navigate to "BaseStagesCompletedStep" instead.
     if (nextStep?.type === 'BaseStagesCompletedStep' && stepAfterNextStep === activeStep) {
-      this.router.transitionTo(nextStep.routeParams.route, nextStep.routeParams.models);
+      this.router.transitionTo(nextStep.routeParams.route, nextStep.routeParams.models[0]!);
```
rohitpaulk, line 85:
```suggestion
      // @ts-expect-error
      this.router.transitionTo(nextStep.routeParams.route, ...nextStep.routeParams.models);
```

`@ts-expect-error` since it looks like Ember's types don't allow variadic arguments yet. `...models` since that's doesn't include an assumption around how many arguments the route has

rohitpaulk (approved):

app/controllers/course/stage/instructions.ts
rohitpaulk: Hmm, actually I'm confused now - how does the invocation above on line 78 work? 

```
this.router.transitionTo(activeStep.routeParams.route, activeStep.routeParams.models);
```

Shouldn't that have this same problem too? 🤔

libmartinito: in the previous line, the route needs two models, course and course stage. tried digging in issues on the ember repo but i couldn't find out why when a route expects a single model that an array of a single identifier doesn't work

PR #1809, opened 2024-05-31
CC-1265 Add code example confirmation modal

libmartinito:
**Checklist**:

## Summary by CodeRabbit

- **New Features**
  - Added a modal to notify users when they attempt to view code examples for an incomplete stage, with options to go back to instructions or view code examples.

- **Tests**
  - Added tests to ensure the stage incomplete modal displays correctly based on the completion status of course stages.

### DIFF
--- a/app/components/course-page/course-stage-step/community-solutions-list.ts
+++ b/app/components/course-page/course-stage-step/community-solutions-list.ts
@@ -11,7 +11,6 @@ type Signature = {
   Element: HTMLDivElement;
 
   Args: {
-    handleStageIncompleteModalDismissed: () => void;
     solutions: CommunityCourseStageSolutionModel[];
     stageIncompleteModalWasDismissed: boolean;
     repository: string;
--- a/app/components/course-page/course-stage-step/community-solutions-list.ts
+++ b/app/components/course-page/course-stage-step/community-solutions-list.ts
@@ -1,10 +1,11 @@
-import rippleSpinnerImage from '/assets/images/icons/ripple-spinner.svg';
 import Component from '@glimmer/component';
-import { tracked } from '@glimmer/tracking';
+import rippleSpinnerImage from '/assets/images/icons/ripple-spinner.svg';
 import { action } from '@ember/object';
 import { service } from '@ember/service';
-import type CommunityCourseStageSolutionModel from 'codecrafters-frontend/models/community-course-stage-solution';
+import { tracked } from '@glimmer/tracking';
 import type AuthenticatorService from 'codecrafters-frontend/services/authenticator';
+import type CommunityCourseStageSolutionModel from 'codecrafters-frontend/models/community-course-stage-solution';
+import type CoursePageStateService from 'codecrafters-frontend/services/course-page-state';
 
 type Signature = {
   Element: HTMLDivElement;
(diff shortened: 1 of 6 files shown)


app/controllers/course/stage/code-examples.ts
```diff
@@ -20,8 +20,9 @@ export default class CodeExamplesController extends Controller {
   rippleSpinnerImage = rippleSpinnerImage;
   @tracked order: 'recommended' | 'recent' | 'affiliated' = 'recommended';
   @tracked isLoading = true;
-  @tracked solutions: CommunityCourseStageSolutionModel[] = [];
   @tracked requestedLanguage: LanguageModel | null = null; // This shouldn't be state on the controller, see if we can move it to a query param or so?
+  @tracked solutions: CommunityCourseStageSolutionModel[] = [];
+  @tracked stageIncompleteModalWasDismissed = false;
```
rohitpaulk, line 25: @libmartinito when this is stored on the controller, it'll persist between different stages. So if a user dismissed the modal on one stage, it'll remain dismissed on other stages too. 

Thoughts on moving this state to `CommunitySolutionList` instead?

app/templates/course/stage/code-examples.hbs
```diff
@@ -58,7 +58,12 @@
       </div>
     {{else if (gt this.sortedSolutions.length 0)}}
       {{! @glint-expect-error: not ts-ified yet }}
```
rohitpaulk, line 60: This is a .ts component now, looks like the comment above is hiding another error. Can we look into what it is?

app/components/course-page/course-stage-step/community-solutions-list.ts
```diff
@@ -19,9 +22,19 @@ export default class CommunitySolutionsListComponent extends Component<Signature
...
 
   @service declare authenticator: AuthenticatorService;
+  @service declare coursePageState: CoursePageStateService;
 
   @tracked configureGithubIntegrationModalIsOpen = false;
 
+  get shouldShowStageIncompleteModal() {
+    return (
+      !this.args.stageIncompleteModalWasDismissed &&
+      this.coursePageState.currentStep.globalPosition > 4 &&
```
rohitpaulk, line 32: Instead of using `globalPosition`, can we filter by `courseStage.isFirst` & `courseStage.isSecond`? `globalPosition` is a bit indirect - if we add another "setup" step for example, the `globalPosition` required here would change from 4 to 5.

rohitpaulk (changes requested):
Notes added

app/templates/course/stage/code-examples.hbs
```diff
@@ -1,4 +1,11 @@
-<div class="pt-8 pb-32" {{did-update this.loadSolutions this.courseStage}} {{did-insert this.loadSolutions}} data-test-code-examples-tab>
+<div
+  class="pt-8 pb-32"
+  {{did-update (fn (mut this.stageIncompleteModalWasDismissed) false) this.courseStage}}
+  {{did-update this.loadSolutions this.courseStage}}
+  {{did-insert this.loadSolutions}}
+  {{will-destroy (fn (mut this.stageIncompleteModalWasDismissed) false) this.courseStage}}
```
rohitpaulk, line 6:
```suggestion
  {{will-destroy (fn (mut this.stageIncompleteModalWasDismissed) false)}}
```

(shouldn't need the extra argument for `will-destroy`)

app/templates/course/stage/code-examples.hbs
```diff
@@ -1,4 +1,11 @@
-<div class="pt-8 pb-32" {{did-update this.loadSolutions this.courseStage}} {{did-insert this.loadSolutions}} data-test-code-examples-tab>
+<div
+  class="pt-8 pb-32"
+  {{did-update (fn (mut this.stageIncompleteModalWasDismissed) false) this.courseStage}}
```
rohitpaulk, line 3: I might be wrong about this, but even if a property on `courseStage` changes, this would fire. (`courseStage.instructionsMarkdown`, for example, or `courseStage.solutionCounts`)

Can we confirm this? In the past, I think I've always ended up storing `courseStage` as a property in the component/controller and then on each callback I'd compare the received courseStage against the stored property and not do anything unless it has changed

rohitpaulk (changes requested):
Needs confirming around the behaviour for `did-update`

rohitpaulk:
Nice work on the extensive tests!

app/templates/course/stage/code-examples.hbs
libmartinito: added another test to check for this

rohitpaulk (approved):

PR #1841, opened 2024-06-10
CC-1285 Allow challenge deprecation in catalog

libmartinito:
**Checklist**:

## Summary by CodeRabbit

- **New Features**
  - Introduced conditional rendering for course cards based on their status.
  - Added handling for deprecated courses in the course catalog.
  - Enhanced course filtering logic to include deprecated status.
  
- **Bug Fixes**
  - Corrected the number of course cards displayed on the catalog page to match the expected count.

- **Tests**
  - Updated tests to validate the display of deprecated courses based on user progress.
  - Increased the number of course cards displayed in test scenarios from 5 to 9.

### DIFF
--- a/app/components/course-card.hbs
+++ b/app/components/course-card.hbs
@@ -1,89 +1,87 @@
-{{#if this.shouldShowCourseCard}}
-  {{! template-lint-disable no-invalid-interactive }}
-  <LinkTo
-    class="group bg-white p-5 rounded-md shadow hover:shadow-md transition-all cursor-pointer flex flex-col justify-between border relative"
-    @route={{this.linkToRoute.name}}
-    @model={{this.linkToRoute.model}}
-    @query={{this.linkToRoute.query}}
-    data-test-course-card={{not this.isSkeleton}}
-  >
-    {{! Logo }}
-    <div class="w-8 transform scale-100 group-hover:scale-105 flex-shrink-0 transition-all absolute top-5 right-5">
+{{! template-lint-disable no-invalid-interactive }}
+<LinkTo
+  class="group bg-white p-5 rounded-md shadow hover:shadow-md transition-all cursor-pointer flex flex-col justify-between border relative"
+  @route={{this.linkToRoute.name}}
+  @model={{this.linkToRoute.model}}
+  @query={{this.linkToRoute.query}}
+  data-test-course-card={{not this.isSkeleton}}
+>
+  {{! Logo }}
+  <div class="w-8 transform scale-100 group-hover:scale-105 flex-shrink-0 transition-all absolute top-5 right-5">
+    {{#if this.isSkeleton}}
+      <div class="bg-gray-100 w-8 h-8 rounded-full" />
+    {{else}}
+      <CourseLogo @course={{@course}} />
+    {{/if}}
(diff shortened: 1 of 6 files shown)


app/components/course-card.hbs
```diff
@@ -1,87 +1,89 @@
...
-        {{#if this.isSkeleton}}
-          <span class="inline-block bg-gray-100 rounded">{{#each (repeat 30)}}&nbsp;{{/each}}</span>
-        {{else}}
-          <span>{{@course.name}}</span>
-        {{/if}}
-      </div>
-    </div>
-
-    <div class="text-sm leading-relaxed text-gray-500 pr-10" data-test-course-description>
+{{#if this.shouldShowCourseCard}}
```
rohitpaulk, line 1: @libmartinito needs to be handled outside, in the caller. I think I've highlighted this before - this style of avoiding rendering isn't ideal because you're "assuming" a use-case that might not be true. Callers can always choose to just not render a component.

The same course-card might be used in other places that have different restrictions. And it makes it far harder to reason about things too. There are two "levels" of filtering now - alpha cards are filtered out by the caller, and then we're further filtering here.

rohitpaulk (dismissed):
Needs re-architecting, not implemented cleanly

app/controllers/catalog.ts
```diff
@@ -103,4 +98,14 @@ export default class CatalogController extends Controller {
       .filterBy('featureIsProductWalkthrough')
       .rejectBy('isDismissed')[0] as FeatureSuggestionModel | null;
   }
+
+  shouldDisplayCourse(course: CourseModel) {
+    const isAlphaOrDeprecated = course.releaseStatusIsAlpha || course.releaseStatusIsDeprecated;
+    const isAuthorizedUser =
```
rohitpaulk, line 104:
```suggestion
    const userIsStaffOrCourseAuthor =
```

app/controllers/catalog.ts
```diff
@@ -103,4 +98,14 @@ export default class CatalogController extends Controller {
...
+      this.authenticator.currentUser && (this.authenticator.currentUser.isStaff || this.authenticator.currentUser.isCourseAuthor(course));
+    const hasUserProgress =
```
rohitpaulk, line 106:
```suggestion
    const userHasRepository =
```

rohitpaulk: (don't introduce "new" names unless they actually simplify logic or compress things, sticking to existing names/concepts keeps things simpler)

app/controllers/catalog.ts
rohitpaulk: (don't introduce "new" names unless they actually simplify logic or compress things, sticking to existing names/concepts keeps things simpler)

rohitpaulk: On second thought, let's change this behaviour - for "alpha" it makes sense to keep a course in the list for authors / staff since they'd need to test a course actively. For "deprecated" this isn't the case, you'd hardly every want to visit a deprecated course. Let's not show deprecated courses even to authors/staff, they can always navigate to the URL (/courses/docker) directly if needed

rohitpaulk (approved):

PR #1867, opened 2024-06-18
CC-1296 add refresh from github button

libmartinito:
**Checklist**:

## Summary by CodeRabbit

- **New Features**
  - Added a tooltip for syncing GitHub account username.
  - Enabled the refresh button functionality to update the username from GitHub.

- **Bug Fixes**
  - Improved GitHub username synchronization mechanism.

- **Tests**
  - Added tests to verify the functionality of refreshing GitHub usernames.

### DIFF
--- a/app/components/settings/profile-page/username-section.hbs
+++ b/app/components/settings/profile-page/username-section.hbs
@@ -26,6 +26,7 @@
     @size="extra-small"
     class="mt-4"
     @shouldShowSpinner={{this.isSyncingUsernameFromGitHub}}
+    @isDisabled={{this.currentUser.hasAnonymousModeEnabled}}
     {{on "click" this.refreshFromGitHub}}
     data-test-refresh-from-github-button
   >
--- a/app/components/settings/profile-page/username-section.hbs
+++ b/app/components/settings/profile-page/username-section.hbs
@@ -22,7 +22,13 @@
     {{/if}}
   </div>
 
-  <TertiaryButtonWithSpinner @size="extra-small" class="mt-4" @shouldShowSpinner={{this.isSyncingGitHubUsername}} {{on "click" this.refreshFromGitHub}} data-test-refresh-from-github-button>
+  <TertiaryButtonWithSpinner
+    @size="extra-small"
+    class="mt-4"
+    @shouldShowSpinner={{this.isSyncingGitHubUsername}}
+    {{on "click" this.refreshFromGitHub}}
+    data-test-refresh-from-github-button
+  >
     <div class="flex items-center gap-1">
       {{svg-jar "refresh" class="w-4 h-4 text-gray-400"}}
       Refresh from GitHub
--- a/app/components/settings/profile-page/username-section.hbs
+++ b/app/components/settings/profile-page/username-section.hbs
@@ -34,5 +34,9 @@
       {{svg-jar "refresh" class="w-4 h-4 text-gray-400"}}
       Refresh from GitHub
     </div>
+
+    {{#if this.currentUser.hasAnonymousModeEnabled}}
+      <EmberTooltip @text="Your profile has anonymous mode enabled. Your GitHub username is not visible to others." />
+    {{/if}}
   </TertiaryButtonWithSpinner>
 </Settings::FormSection>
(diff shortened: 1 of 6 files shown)


app/components/settings/profile-page/username-section.ts
```diff
@@ -9,7 +13,18 @@ interface Signature {
...
-export default class UsernameSectionComponent extends Component<Signature> {}
+export default class UsernameSectionComponent extends Component<Signature> {
+  @service declare authenticator: AuthenticatorService;
+
+  @tracked isSyncingGitHubUsername = false;
+
+  @action
+  async refreshFromGitHub() {
+    this.isSyncingGitHubUsername = true;
+    await this.authenticator.currentUser?.syncGitHubUsername(null);
```
libmartinito, line 24: do you think it's better to keep the 'refresh from github' name consistent everywhere

tests/acceptance/settings-page/profile-test.js
```diff
@@ -47,4 +48,19 @@ module('Acceptance | settings-page | profile-test', function (hooks) {
...
+    testScenario(this.server);
+    signInAsSubscriber(this.owner, this.server);
+
+    let currentUser = this.server.schema.users.first();
+
+    assert.strictEqual(currentUser.username, 'rohitpaulk');
+
+    await profilePage.visit();
+    await profilePage.refreshFromGitHubButton.click();
+    await settled();
```
rohitpaulk, line 62: This shouldn't be required, `click()` should already be calling `await settled()`. @libmartinito are you sure that this fails without a call to `settled()`?

app/components/settings/profile-page/username-section.ts
rohitpaulk: `sync` is fine, we use it across many `core` actions so it's consistent with those. The actual text for the button might change over time (based on what makes most sense to a user), so not super important to keep that in sync with everything else. 

Would recommend renaming to `syncUsernameFromGitHub` though, to avoid confusion around whether we're actually updating the username, or only `user#githubUsername`.

app/components/settings/profile-page/username-section.hbs
```diff
@@ -11,24 +11,27 @@
...
       Your profile has anonymous mode enabled. Your GitHub username is not visible to others.
     {{else}}
-      This value is synced with your GitHub account.
+      This value is synced with your GitHub account. Click on the refresh button below to update your username.
     {{/if}}
   </div>
 
-  {{!-- TODO: Bring this back once implemented }}
-  {{!-- <TertiaryButtonWithSpinner @size="extra-small" class="mt-4" @shouldShowSpinner={{false}}>
+  <TertiaryButtonWithSpinner
```
rohitpaulk, line 25: If a user has anonymous mode enabled, should we show this?

rohitpaulk (changes requested):
Added q

app/components/settings/profile-page/username-section.hbs
rohitpaulk: If a user has anonymous mode enabled and clicked on this, what would happen?

tests/acceptance/settings-page/profile-test.js
libmartinito: yep, works without the additional call to settled

app/components/settings/profile-page/username-section.hbs
libmartinito: makes sense. user shouldn't be able to change github username if anonymous

app/components/settings/profile-page/username-section.hbs
```diff
@@ -11,24 +11,28 @@
...
+    @size="extra-small"
+    class="mt-4"
+    @shouldShowSpinner={{this.isSyncingUsernameFromGitHub}}
+    @isDisabled={{this.currentUser.hasAnonymousModeEnabled}}
```
rohitpaulk, line 29: @libmartinito let's add a tooltip when this button is disabled: 

```
Your profile has anonymous mode enabled. Your GitHub username is not visible to others.
```

rohitpaulk (approved):

PR #1872, opened 2024-06-18
CC-1284 Add preventScroll: true to all focus method calls on an html element

libmartinito:
**Checklist**:

## Summary by CodeRabbit

- **New Features**
	- Enhanced focus behavior across various components to prevent unwanted scrolling when elements are focused.
		- Applied changes to `QuestionCardComponent`, `PrimaryButton`, `RequestLanguageDropdownComponent`, `CancelSubscriptionModalComponent`, and `FocusOnInsertModifier`.

### DIFF
--- a/app/components/concept-page/question-card.ts
+++ b/app/components/concept-page/question-card.ts
@@ -57,7 +57,7 @@ export default class QuestionCardComponent extends Component<Signature> {
     if (firstOptionElement instanceof HTMLElement) {
       // focus() doesn't seem to work unless it's called after the current runloop
       next(() => {
-        firstOptionElement && firstOptionElement.focus();
+        firstOptionElement && firstOptionElement.focus({ preventScroll: true });
       });
     }
   }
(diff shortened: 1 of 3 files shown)


app/components/course-page/introduction-step/create-repository-card/request-language-dropdown.js
```diff
@@ -68,7 +68,7 @@ export default class RequestLanguageDropdownComponent extends Component {
   @action
   handleInputDidInsert(element) {
     this.inputElement = element;
-    this.inputElement.focus();
+    this.inputElement.focus({ preventScroll: true });
```
rohitpaulk, line 71: @libmartinito I'm not sure about changing this everywhere else - it makes sense for concepts since we **know** that scrolling is handled somewhere else. In these other cases I'm not sure if this is the case. Can we keep the other use-cases as-is for now? If we discover more scroll+focus interactions then we can think about how to handle them separately. Even for `focus-on-insert`, `preventScroll` should be an option - there are places where we'd want this and places where we wouldn't.

rohitpaulk (dismissed):
Let's restrict changes to only concepts, not sure about whether `preventScroll` needs to be a "universal" think yet

rohitpaulk (approved):

PR #1892, opened 2024-06-24
CC-1325 Fix scroll bars showing by default

libmartinito:
**Checklist**:

## Summary by CodeRabbit

- **Style**
  - Updated various components and templates to change CSS classes from `overflow-scroll` to `overflow-auto`, improving the overflow behavior for both horizontal and vertical scrolling. This change enhances the user interface by adjusting the way content scrolls within different elements.


libmartinito:
is it fine to replace overflow-scroll with overflow-auto everywhere? not sure if there are cases where we always want to show the space for the scrollbar

rohitpaulk (approved):

PR #1898, opened 2024-06-25
CC-981 Support toggling concept status

libmartinito:
**Checklist**:

## Summary by CodeRabbit

- **New Features**
  - Introduced a "Draft" label with a tooltip for draft concepts visible only to staff.
  - Added a toggle switch to control concept publishing status from the "Basic Details" admin page.

- **Enhancements**
  - Improved concept visibility filtering to show only relevant concepts based on user roles and statuses.

- **Tests**
  - Added test cases to verify the correct toggling of concept publish status and the visibility of draft labels for both draft and published concepts.

### DIFF
--- a/app/components/concepts-page/concept-card.hbs
+++ b/app/components/concepts-page/concept-card.hbs
@@ -13,6 +13,13 @@
       {{#if (current-user-is-concept-author @concept)}}
         <Pill @color="green">Yours</Pill>
       {{/if}}
+
+      {{#if (eq @concept.status "draft")}}
+        <Pill @color="gray">
+          Draft
+          <EmberTooltip @text="This concept is only visible to staff" />
+        </Pill>
+      {{/if}}
     </div>
 
     <div class="prose prose-sm mb-5">
--- a/app/components/concepts-page/concept-card.hbs
+++ b/app/components/concepts-page/concept-card.hbs
@@ -15,7 +15,7 @@
       {{/if}}
 
       {{#if (eq @concept.status "draft")}}
-        <Pill @color="gray">
+        <Pill @color="gray" data-test-draft-label>
           Draft
           <EmberTooltip @text="This concept is only visible to staff" />
         </Pill>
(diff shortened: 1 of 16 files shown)


app/components/concepts-page/concept-card.hbs
```diff
@@ -13,6 +13,13 @@
       {{#if (current-user-is-concept-author @concept)}}
         <Pill @color="green">Yours</Pill>
       {{/if}}
+
+      {{#if (eq @concept.status "draft")}}
+        <Pill @color="gray" data-test-draft-label>
+          Draft
+          <EmberTooltip @text="This concept is only visible to staff" />
+        </Pill>
+      {{/if}}
```
libmartinito, line 22: feels like the two pills communicate something different, so added both

app/controllers/concepts.ts
```diff
@@ -18,12 +19,24 @@ export default class ConceptsController extends Controller {
   @service declare router: RouterService;
   @service declare authenticator: AuthenticatorService;
 
+  get currentUser(): UserModel | null {
```
rohitpaulk, line 22: If the return type says `UserModel | null`, why are we returning with `as UserModel`?

app/templates/concept-admin/basic-details.hbs
```diff
@@ -103,6 +103,28 @@
...
+            Unpublished concepts aren't visible to other users.
+          </label>
+        </:description>
+
+        <:content>
+          <Toggle @isOn={{this.isConceptPublished}} {{on "click" this.handlePublishConceptToggled}} class="mt-1" data-test-publish-concept-toggle />
+
+          {{#if this.isConceptPublished}}
+            <div class="text-gray-400 text-xs mt-0.5">
+              This concept is published
```
rohitpaulk, line 119: If a concept was `published` already, this would show a wrong value

rohitpaulk (changes requested):
Added notes

rohitpaulk (approved):

PR #2365, opened 2024-10-28
Unify all interactions with Clipboard & stub all Clipboard methods during tests

VasylMarchuk:
### Brief

This unifies all interactions with the Clipboard by adding a new component `CopyableAnything`, and by using it in all components that previously interacted with `navigator.clipboard` manually (custom implementations in approximately 9 different places, each slightly different).

In addition, when running the tests suite, all `navigator.clipboard.*` API methods are stubbed with Sinon fakes before each test, and un-stubbed after, thus preventing any potential side-effects.

The timeout of the "Copied!" message shown after copying is also reduced from `1000` to `10` milliseconds in the `test` environment.

### Details

- Added `CopyableAnything` component:
  - `@value` value to copy to the clipboard
  - `@onCopied` callback to call after copying
  - `@isDisabled` when `true`, calling yielded `performCopyTask` does nothing
  - yields `performCopyTask` — call this to actually copy passed `@value` to clipboard
  - yields `hasRecentlyCopied` — will be `true` for a configured time after copying
  - tests included
- Rewrote all components that used `navigator.clipboard.writeText` to use `CopyableAnything` component instead:
  - `CopyableCode`
  - `CopyableTerminalCommand`
  - `ComparisonCard`
  - `EvaluationTab`
  - `HeaderContainer`
  - `RepositoryDropdown`
  - `ShareProgressModal`
- Added `stubClipboard` test helper:
  - runs before each test
  - no side-effects when calling `navigator.clipboard.writeText` in `test` environment
  - no need to check `if (config.environment === 'test')` in the code
- Added `copyConfirmationTimeout: 1000` to config, set it to `10` in in `test` environment
- Extracted `localStorage`-stubbing logic into a test helper `stubLocalStorage`

**Checklist**:

## Release Notes

- **New Features**
	- Introduced the `CopyableAnything` component for enhanced clipboard functionality across various components.
	- Added a configuration property `copyConfirmationTimeout` for customizable copy feedback duration.

- **Bug Fixes**
	- Improved user feedback for copy actions by dynamically updating button text based on copy status.

- **Refactor**
	- Removed redundant copy handling methods from several components, centralizing functionality in `CopyableAnything`.

- **Tests**
	- Added integration tests for the `CopyableAnything` component to validate copy functionality and state management.
	- Introduced stubbing functions for clipboard and local storage to streamline testing processes.


rohitpaulk:
Dismissing review, since our plan was to make a more minimal changeset

PR #2367, opened 2024-10-29
CC-1480: Highlight language guides in stage 2 tutorial card

andy1li:
**Checklist**:

## Summary by CodeRabbit

- **New Features**
	- Introduced dynamic visibility for language guides in the instructions card, enhancing user experience.
	- Added conditional rendering for language guide recommendations in the "Implement Solution" and "Read Instructions" steps.
	- Enhanced responsiveness of the course stage instructions by integrating language guide loading.
	- Updated user interface to display language-specific guidance based on the availability of language guides.

- **Bug Fixes**
	- Improved state management for displaying language guides based on feature flags.

- **Documentation**
	- Updated component interfaces to include new properties related to language guide visibility.

### DIFF
--- a/app/components/course-page/course-stage-step/second-stage-instructions-card.hbs
+++ b/app/components/course-page/course-stage-step/second-stage-instructions-card.hbs
@@ -1,10 +1,4 @@
-<CoursePage::InstructionsCard
-  @title="How to pass this stage"
-  id="second-stage-instructions-card"
-  ...attributes
-  {{did-update this.loadLanguageGuides @courseStage}}
-  {{did-insert this.loadLanguageGuides}}
->
+<CoursePage::InstructionsCard @title="How to pass this stage" id="second-stage-instructions-card" ...attributes>
   <:content>
     <div class="prose dark:prose-invert mb-5">
       <p>
@@ -24,14 +18,14 @@
           @repository={{@repository}}
           @courseStage={{@courseStage}}
           @isComplete={{this.readInstructionsStepIsComplete}}
-          @isShowingLanguageGuide={{this.isShowingLanguageGuide}}
+          @shouldRecommendLanguageGuide={{@shouldRecommendLanguageGuide}}
         />
       {{else if (eq stepList.expandedStep.id "implement-solution")}}
         <CoursePage::CourseStageStep::SecondStageInstructionsCard::ImplementSolutionStep
           @repository={{@repository}}
           @courseStage={{@courseStage}}
           @isComplete={{this.implementSolutionStepIsComplete}}
-          @isShowingLanguageGuide={{this.isShowingLanguageGuide}}
+          @shouldRecommendLanguageGuide={{@shouldRecommendLanguageGuide}}
         />
       {{else if (eq stepList.expandedStep.id "run-tests")}}
         <CoursePage::CourseStageStep::SecondStageInstructionsCard::RunTestsStep
--- a/app/components/course-page/course-stage-step/second-stage-instructions-card.hbs
+++ b/app/components/course-page/course-stage-step/second-stage-instructions-card.hbs
@@ -1,4 +1,10 @@
-<CoursePage::InstructionsCard @title="How to pass this stage" id="second-stage-instructions-card" ...attributes>
+<CoursePage::InstructionsCard
+  @title="How to pass this stage"
+  id="second-stage-instructions-card"
+  ...attributes
+  {{did-update this.loadLanguageGuides @courseStage}}
+  {{did-insert this.loadLanguageGuides}}
+>
   <:content>
     <div class="prose dark:prose-invert mb-5">
       <p>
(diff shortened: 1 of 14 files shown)


app/components/course-page/course-stage-step/second-stage-instructions-card.hbs
```diff
@@ -1,4 +1,10 @@
-<CoursePage::InstructionsCard @title="How to pass this stage" id="second-stage-instructions-card" ...attributes>
+<CoursePage::InstructionsCard
+  @title="How to pass this stage"
+  id="second-stage-instructions-card"
+  ...attributes
+  {{did-update this.loadLanguageGuides @courseStage}}
```
rohitpaulk, line 5: @andy1li this means that two components will be individually trying to load language guides on the same page. Can we move this into the `instructions.ts` controller please? Once it's loaded there, other components can just use methods that do something similar to what we do for `solutions`: 

Screenshot 2024-10-29 at 11 31 32

rohitpaulk (changes requested):
Note added

tests/acceptance/course-page/try-other-language-test.js
```diff
@@ -36,7 +36,7 @@ module('Acceptance | course-page | try-other-language', function (hooks) {
       'fetch repositories (course page)',
       'fetch leaderboard entries (course page)',
       'fetch hints (course page)',
-      // 'fetch language guide (course page)',
+      'fetch language guide (course page)',
```
rohitpaulk, line 39: Hmm, weren't we already loading this from the current card on the page? 🤔 @andy1li do you know why this only has to be uncommented now?

app/components/course-page/course-stage-step/second-stage-instructions-card/implement-solution-step.ts
```diff
@@ -9,10 +9,15 @@ interface Signature {
...
     courseStage: CourseStageModel;
     isComplete: boolean;
+    isShowingLanguageGuide: boolean;
   };
 }
 
-export default class ImplementSolutionStepComponent extends Component<Signature> {}
+export default class ImplementSolutionStepComponent extends Component<Signature> {
+  get currentLanguageName(): string {
+    return this.args.repository.language!.name;
```
rohitpaulk, line 18: @andy1li this level of indirection usually isn't worth it - can use `@repository.language.name` directly in the template 🙂 Also, I think that this card _might_ render when a language isn't selected, so in that case using the template method won't throw an exception (templates tolerate null values and just render empty contents, vs. this method would actually through an exception)

tests/acceptance/course-page/try-other-language-test.js
andy1li: @rohitpaulk I needed to uncomment three of them, otherwise the tests won't pass.

Context:

These tests were set up on the second stage of redis. It seems that language guides had been loaded before recent changes.

app/components/course-page/course-stage-step/second-stage-instructions-card/read-instructions-step.hbs
```diff
@@ -9,9 +9,16 @@
   </p>
   <ul>
     <li>
-      The
-      <LinkTo @route="course.stage.code-examples">Code Examples</LinkTo>
-      tab, which contains code examples from other users.
+      {{#if @isShowingLanguageGuide}}
+        The
+        <a href="#language-guide-card">{{this.currentLanguageName}} Guide</a>
```
rohitpaulk, line 14:
```suggestion
        <a href="#language-guide-card">"{{this.currentLanguageName}} Guide"</a>
```

(consistent with the "Your Task" quoting above)

andy1li: @andy1li Should I fix these too? (It seems that "Your Task" is the odd one here.)

image

rohitpaulk: > It seems that "Your Task" is the odd one here.

Ah, nice catch - yeah okay let's remove the quotes from Your Task then! Could look weird to have a lot of quoted stuff there.

app/controllers/course/stage/instructions.ts
```diff
@@ -47,7 +50,13 @@ export default class CourseStageInstructionsController extends Controller {
   }
 
   get shouldShowLanguageGuide() {
-    return !this.model.courseStage.isFirst && (this.featureFlags.canSeeLanguageGuidesForStage2 || this.authenticator.currentUser?.isStaff);
+    const languageGuide = this.model.courseStage.languageGuides.findBy('language', this.model.activeRepository.language);
+
+    return (
+      !this.model.courseStage.isFirst &&
+      !!languageGuide &&
+      (this.featureFlags.canSeeLanguageGuidesForStage2 || this.authenticator.currentUser?.isStaff)
```
rohitpaulk, line 58:
```suggestion
      this.featureFlags.canSeeLanguageGuidesForStage2
```

(We already have the staff check inside `canSeeLanguageGuidesForStage2`)

rohitpaulk (approved):
Code looks good! Will briefly discuss this on our call before releasing

app/components/course-page/course-stage-step/simple-language-guide-card.ts
```diff
@@ -18,8 +18,8 @@ interface Signature {
 export default class SimpleLanguageGuideCardComponent extends Component<Signature> {
   @service declare store: Store;
 
-  get languageGuide() {
-    return this.args.courseStage.languageGuides.findBy('language', this.args.language);
+  get languageGuide(): CourseStageLanguageGuideModel {
```
rohitpaulk, line 21: Change component signature, to accept `courseStageLanguageGuide` instead.

PR #2388, opened 2024-11-06
Fix resume track button

andy1li:
**Checklist**:

## Summary by CodeRabbit

- **New Features**
	- Enhanced routing functionality for the Resume Track button, allowing the inclusion of query parameters when transitioning to the course route.
	- Added visibility checks and interaction capabilities for the "resume track" button based on user authentication and course progress.

- **Bug Fixes**
	- Improved handling of course tracking by ensuring the correct parameters are passed during navigation.

- **Tests**
	- Introduced new acceptance tests for the track page functionality, verifying the behavior of the "resume track" button for different user states and course progress.
	- Updated test organization for better clarity and structure in acceptance tests related to the track page.
	- Enhanced page object functionality with new properties for interacting with and checking the visibility of the "resume track" button.

### DIFF
--- a/app/components/track-page/resume-track-button.ts
+++ b/app/components/track-page/resume-track-button.ts
@@ -27,10 +27,9 @@ export default class CourseOverviewResumeTrackButtonComponent extends Component<
     }
 
     return this.args.courses.find((course) => {
-      return !this.authenticator
-        .currentUser!.repositories.filterBy('course', course)
-        .filterBy('language', this.args.language)
-        .some((repository) => repository.allStagesAreComplete);
+      const courseRepositories = this.authenticator.currentUser!.repositories.filterBy('course', course).filterBy('language', this.args.language);
+
+      return courseRepositories.length > 0 && !courseRepositories.some((repository) => repository.allStagesAreComplete);
     });
   }
 
--- a/app/components/track-page/resume-track-button.ts
+++ b/app/components/track-page/resume-track-button.ts
@@ -36,7 +36,9 @@ export default class CourseOverviewResumeTrackButtonComponent extends Component<
 
   @action
   handleClicked() {
-    this.router.transitionTo('course', (this.activeCourse || this.args.courses.reverse()[0]!).slug);
+    this.router.transitionTo('course', (this.activeCourse || this.args.courses.reverse()[0]!).slug, {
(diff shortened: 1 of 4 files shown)


rohitpaulk:
@andy1li let's add a test for this? This is a pretty common user flow, so I think it's worth adding a test to cover this.

app/components/track-page/resume-track-button.ts
```diff
@@ -27,16 +27,17 @@ export default class CourseOverviewResumeTrackButtonComponent extends Component<
     }
 
     return this.args.courses.find((course) => {
-      return !this.authenticator
-        .currentUser!.repositories.filterBy('course', course)
-        .filterBy('language', this.args.language)
-        .some((repository) => repository.allStagesAreComplete);
+      const courseRepositories = this.authenticator.currentUser!.repositories.filterBy('course', course).filterBy('language', this.args.language);
+
+      return courseRepositories.length > 0 && !courseRepositories.some((repository) => repository.allStagesAreComplete);
```
andy1li, line 32: There's a subtle bug for course with zero repos:
- `![].some` is true, 
- but we don't want to resume a course with zero repos.

tests/pages/track-page.js
```diff
@@ -12,6 +12,9 @@ export default createPage({
     await finishRender(); // Page has poller
   },
 
+  clickResumeTrackButton: clickable('[data-test-resume-track-button]'),
```
rohitpaulk, line 15:
```suggestion
  clickOnResumeTrackButton: clickable('[data-test-resume-track-button]'),
```

(consistent)

app/components/track-page/resume-track-button.ts
rohitpaulk: @andy1li Can you clarify this? `[].some()` shouldn't return true if there are no elements (as far as I know).

rohitpaulk (changes requested):
Question added

app/components/track-page/resume-track-button.ts
rohitpaulk: Oh... k got it, `some` returns false, but there are zero repos. Cool!

rohitpaulk (approved):

PR #2392, opened 2024-11-08
Stub Clipboard methods during tests

VasylMarchuk:
Extracted from https://github.com/codecrafters-io/frontend/pull/2365

### Brief

Stub all clipboard methods when running tests using Sinon's fakes.

### Details

- Added `stubClipboard` test helper and made run it before each test
  - it stubs all clipboard methods with Sinon fakes in accordance with the [MDN API Docs](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard)
  - all fakes are restored back to original methods after each test 
- Removed all `environment === 'test'` checks from component methods calling `navigator.clipboard.writeText`
- Added `copyConfirmationTimeout: 1000` to config
  - made all components use it instead of hardcoded value
  - during tests, `copyConfirmationTimeout` is set to `10`
- Added missing `await` statements to all `navigator.clipboard.writeText` calls
- Bonus: moved `localStorage`-stubbing logic into a test helper `stubLocalStorage` as well
  - much cleaner main `test-helper.js` file
  - rewritten `stubLocalStorage` in TypeScript
  - made sure it's compliant with [MDN API Docs](https://developer.mozilla.org/en-US/docs/Web/API/Storage)
- Both `stubClipboard` and `stubLocalStorage` test helpers can be called multiple times during a test
  - they also return the underlying cache (`Array` or `Map`) used to store temporary fake values
  - these arrays can be manipulated at wish, if needed for some tests

### Checklist

## Summary by CodeRabbit

- **New Features**
	- Enhanced clipboard functionality across multiple components with asynchronous handling.
	- Dynamic timeout duration for clipboard operations based on the environment.

- **Bug Fixes**
	- Corrected a typographical error in the event handler name for copying IDs to the clipboard.

- **Tests**
	- Introduced helper functions for stubbing clipboard and local storage methods to streamline testing.

### DIFF
--- a/app/components/copyable-code.ts
+++ b/app/components/copyable-code.ts
@@ -2,6 +2,7 @@ import Component from '@glimmer/component';
 import { action } from '@ember/object';
 import { later } from '@ember/runloop';
 import { tracked } from '@glimmer/tracking';
+import config from 'codecrafters-frontend/config/environment';
 
 interface Signature {
   Element: HTMLDivElement;
@@ -27,7 +28,7 @@ export default class CopyableCodeComponent extends Component<Signature> {
       () => {
         this.codeWasCopiedRecently = false;
       },
-      1000,
+      config.x.copyConfirmationTimeout,
     );
 
     if (this.args.onCopyButtonClick) {
--- a/app/components/copyable-code.ts
+++ b/app/components/copyable-code.ts
@@ -28,7 +28,7 @@ export default class CopyableCodeComponent extends Component<Signature> {
       () => {
         this.codeWasCopiedRecently = false;
       },
-      config.x.copyConfirmationTimeout,
+      config.environment !== 'test' ? 1000 : 10,
     );
 
     if (this.args.onCopyButtonClick) {
--- a/app/components/copyable-code.ts
+++ b/app/components/copyable-code.ts
@@ -18,8 +18,8 @@ export default class CopyableCodeComponent extends Component<Signature> {
   @tracked codeWasCopiedRecently: boolean = false;
 
   @action
-  handleCopyButtonClick() {
-    navigator.clipboard.writeText(this.args.code);
+  async handleCopyButtonClick() {
+    await navigator.clipboard.writeText(this.args.code);
 
     this.codeWasCopiedRecently = true;
 
--- a/app/components/copyable-code.ts
+++ b/app/components/copyable-code.ts
@@ -1,5 +1,4 @@
 import Component from '@glimmer/component';
-import config from 'codecrafters-frontend/config/environment';
 import { action } from '@ember/object';
 import { later } from '@ember/runloop';
 import { tracked } from '@glimmer/tracking';
@@ -19,9 +18,7 @@ export default class CopyableCodeComponent extends Component<Signature> {
 
   @action
   handleCopyButtonClick() {
-    if (config.environment !== 'test') {
-      navigator.clipboard.writeText(this.args.code);
-    }
+    navigator.clipboard.writeText(this.args.code);
 
     this.codeWasCopiedRecently = true;
 
--- a/app/components/copyable-terminal-command.ts
+++ b/app/components/copyable-terminal-command.ts
@@ -5,6 +5,7 @@ import { tracked } from '@glimmer/tracking';
 import fade from 'ember-animated/transitions/fade';
 import type DarkModeService from 'codecrafters-frontend/services/dark-mode';
 import { service } from '@ember/service';
+import config from 'codecrafters-frontend/config/environment';
 
 interface Signature {
   Element: HTMLDivElement;
@@ -44,7 +45,7 @@ export default class CopyableTerminalCommandComponent extends Component<Signatur
       () => {
         this.wasCopiedRecently = false;
       },
-      1000,
+      config.x.copyConfirmationTimeout,
     );
 
     if (this.args.onCopyButtonClick) {
--- a/app/components/copyable-terminal-command.ts
+++ b/app/components/copyable-terminal-command.ts
@@ -45,7 +45,7 @@ export default class CopyableTerminalCommandComponent extends Component<Signatur
       () => {
         this.wasCopiedRecently = false;
       },
-      config.x.copyConfirmationTimeout,
+      config.environment !== 'test' ? 1000 : 10,
     );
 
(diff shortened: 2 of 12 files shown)


tests/support/stub-local-storage.ts
```diff
@@ -0,0 +1,45 @@
...
+    cache.clear();
+  });
+
+  stub(window.localStorage, 'key').callsFake(function (index) {
+    return [...cache.keys()][index] || null;
+  });
+
+  stub(window.localStorage['__proto__'], 'length').get(function () {
+    return cache.size;
+  });
```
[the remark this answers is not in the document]
VasylMarchuk: I need to pass the object for Sinon to stub the specified method. Can't use `Object.defineProperty`, I don't define the property, Sinon does.

tests/support/stub-local-storage.ts
```diff
@@ -0,0 +1,45 @@
...
+    if (cache.has(key)) {
+      return String(cache.get(key));
+    } else {
+      return null;
+    }
+  });
+
+  stub(window.localStorage, 'setItem').callsFake(function (key, value) {
+    cache.set(key, value);
+  });
```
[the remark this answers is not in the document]
VasylMarchuk: Irrelevant, same comment as for Clipboard, we're just imitating the behavior and not actually calling the native methods.

app/components/course-admin/code-example-page/comparison-card.ts
```diff
@@ -69,8 +69,8 @@
   }
 
   @action
-  handleCopyIdToClipbardButtonClick() {
-    navigator.clipboard.writeText(this.args.comparison.id);
+  async handleCopyIdToClipbardButtonClick() {
+    await navigator.clipboard.writeText(this.args.comparison.id);
```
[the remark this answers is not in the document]
VasylMarchuk: Irrelevant for this PR

app/components/copyable-code.ts
```diff
@@ -30,7 +28,7 @@ export default class CopyableCodeComponent extends Component<Signature> {
       () => {
         this.codeWasCopiedRecently = false;
       },
-      1000,
+      config.x.copyConfirmationTimeout,
```
rohitpaulk, line 31: Hmm, not sure about using a single value for this - even though we're okay with a (small) fixed value for tests, for production, it's likely that in different contexts this'd need to be different. 

For example, if this was a secret link / password, we'd choose to just keep the confirmation timeout alive for longer (to force the user to take the next action, and not distract) 

Maybe we could just do something like `env === 'test' ? 10 : 1000` here?

app/components/copyable-terminal-command.ts
```diff
@@ -35,10 +35,8 @@ export default class CopyableTerminalCommandComponent extends Component<Signatur
   }
 
   @action
-  handleCopyButtonClick() {
-    if (config.environment !== 'test') {
-      navigator.clipboard.writeText(this.copyableText);
-    }
+  async handleCopyButtonClick() {
+    await navigator.clipboard.writeText(this.copyableText);
```
rohitpaulk, line 39: Nice!

app/components/copyable-terminal-command.ts
```diff
@@ -47,7 +45,7 @@ export default class CopyableTerminalCommandComponent extends Component<Signatur
       () => {
         this.wasCopiedRecently = false;
       },
-      1000,
+      config.x.copyConfirmationTimeout,
```
rohitpaulk, line 48: Same comment as above

rohitpaulk (approved):
Looks good other than the config extraction!

app/components/copyable-code.ts
VasylMarchuk: I thought that according to testing "conventions" the code tested should not be aware of the fact that it's being tested, meaning that `env === "tested"` is considered a very un-orthodox to do?

How about adding a second configuration variable, `copyConfirmationTimeoutImportantMessage: 10000`, for cases where you'd need a longer confirmation? I don't think this would need a TON of variations... maybe short/medium/long at maximum.

VasylMarchuk: On a second thought, `env === 'test' ? 10 : 1000` is not as evil as 

```js
if (config.environment !== 'test') {
  navigator.clipboard.writeText(this.args.code);
}
```

because the former just changes the timeout, while the latter skips an API invocation completely 😱

VasylMarchuk: But the counter-argument is that there already like 8 places where this timeout is hard-coded in various places, and each of them would need to include the env check, and I'm sure someone will forget the env check when he adds a ninth place :P 

I thought about adding a `static confirmationTimeout` fields to the components themselves, but they are a PITA to override in tests. Much easier to have them in config.

VasylMarchuk: So, your call, @rohitpaulk ! :D 

Centralize them into config vars or not?

rohitpaulk: > because the former just changes the timeout, while the latter skips an API invocation completely

Yep - agree with this, the latter is indeed error prone. 

@VasylMarchuk I'd just use `env === 'test' ? 10 : 1000` where necessary, doesn't seem like that'd cause too many issues

VasylMarchuk: Perfect, I don't really have a strong opinion on this, just wanted to have some else's to confirm, feel perfectly fine with `env === 'test' ? 10 : 10000`

app/components/course-page/share-progress-modal.ts
```diff
@@ -64,12 +64,9 @@ export default class ShareProgressModalComponent extends Component<Signature> {
...
   copyToClipboardAndFlashMessage = task({ keepLatest: true }, async (): Promise<void> => {
-    if (config.environment !== 'test') {
-      await navigator.clipboard.writeText(this.copyableText);
-    }
-
+    await navigator.clipboard.writeText(this.copyableText);
     this.wasCopiedRecently = true;
-    await timeout(1000);
+    await timeout(config.environment !== 'test' ? 1000 : 10);
     this.wasCopiedRecently = false;
```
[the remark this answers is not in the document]
VasylMarchuk: We aren't handling clipboard errors in this PR, and in addition, it looks like we track the user's _clicking on the "Copy" button_, not the fact whether it succeeded or not.

PR #2393, opened 2024-11-11
Prepare for CC-1501: Ship experiment with Diff in stage 2 tutorial card

andy1li:
**Checklist**:

- [x] I've added tests for my changes

## Summary by CodeRabbit

- **New Features**
	- Introduced a new computed property to control the visibility of solutions based on course stage.
	- Added properties to conditionally display a solution prompt within the second stage instructions card.
	- Implemented a new action to toggle the visibility of the solution.
	- Added a new method to retrieve the solution for the second stage of a course.
	- Enhanced interactivity with new clickable actions for revealing solutions and checking UI element visibility.

- **Bug Fixes**
	- Removed outdated feature flag dependencies, enhancing the reliability of UI element visibility.

- **Refactor**
	- Streamlined feature flag logic for better clarity and maintainability.

### DIFF
--- a/app/components/course-page/course-stage-step/second-stage-instructions-card/implement-solution-step.hbs
+++ b/app/components/course-page/course-stage-step/second-stage-instructions-card/implement-solution-step.hbs
@@ -2,6 +2,7 @@
   <p>
     Head over to your editor / IDE and implement your solution.
   </p>
+
   <p>
     {{#if @shouldRecommendLanguageGuide}}
       If you want a quick look at what functions to use or how to structure your code, we recommend looking at
@@ -16,3 +17,35 @@
     {{/if}}
   </p>
 </div>
+
+{{#if (and @shouldShowSolution this.solution)}}
+  <BlurredOverlay @isBlurred={{this.solutionIsBlurred}}>
+    <:content>
+      <div class="flex flex-col gap-4 my-3">
+        {{#each this.solution.changedFiles as |changedFile|}}
+          {{! Extra if condition convinces typescript that solution isn't null }}
+          {{#if this.solution}}
+            <FileDiffCard @filename={{changedFile.filename}} @code={{changedFile.diff}} @language={{this.solution.language.slug}} />
+          {{/if}}
+        {{/each}}
+      </div>
+
+      {{#if @shouldRecommendLanguageGuide}}
+        <p class="prose dark:prose-invert prose-compact">
+          For a more detailed explanation on how this solution works, view the
+          <a href="#language-guide-card">{{@repository.language.name}} Guide</a>
+          card.
+        </p>
+      {{/if}}
+    </:content>
+
+    <:overlay>
+      <PrimaryButton {{on "click" this.toggleSolution}} class="self-center">
+        <div class="flex items-center gap-2">
+          {{svg-jar "eye" class="size-6"}}
+          <span>Reveal Solution</span>
+        </div>
+      </PrimaryButton>
+    </:overlay>
+  </BlurredOverlay>
+{{/if}}
--- a/app/components/course-page/course-stage-step/second-stage-instructions-card/implement-solution-step.hbs
+++ b/app/components/course-page/course-stage-step/second-stage-instructions-card/implement-solution-step.hbs
@@ -25,7 +25,12 @@
         {{#each this.solution.changedFiles as |changedFile|}}
           {{! Extra if condition convinces typescript that solution isn't null }}
           {{#if this.solution}}
-            <FileDiffCard @filename={{changedFile.filename}} @code={{changedFile.diff}} @language={{this.solution.language.slug}} />
+            <FileDiffCard
+              @filename={{changedFile.filename}}
+              @code={{changedFile.diff}}
+              @language={{this.solution.language.slug}}
+              data-test-file-diff-card
+            />
           {{/if}}
         {{/each}}
       </div>
@@ -40,7 +45,7 @@
     </:content>
 
     <:overlay>
-      <PrimaryButton {{on "click" this.toggleSolution}} class="self-center">
+      <PrimaryButton {{on "click" this.handleRevealSolutionButtonClick}} class="self-center" data-test-reveal-solution-button>
         <div class="flex items-center gap-2">
           {{svg-jar "eye" class="size-6"}}
           <span>Reveal Solution</span>
--- a/app/controllers/course/stage/instructions.ts
+++ b/app/controllers/course/stage/instructions.ts
@@ -61,8 +61,8 @@ export default class CourseStageInstructionsController extends Controller {
     return !!this.prerequisiteInstructionsMarkdown;
   }
 
-  get shouldShowSolution() {
-    return this.featureFlags.canSeeSolutionsForStage2;
+  get shouldShowStage2Solution() {
+    return this.featureFlags.canSeeSolutionsForStage2 && !!this.model.activeRepository.secondStageSolution;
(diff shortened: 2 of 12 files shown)


app/components/course-page/course-stage-step/second-stage-instructions-card/implement-solution-step.hbs
```diff
@@ -15,4 +16,36 @@
...
+        <p class="prose dark:prose-invert prose-compact">
+          For a more detailed explanation on how this solution works, view the
+          <a href="#language-guide-card">{{@repository.language.name}} Guide</a>
+          card.
+        </p>
+      {{/if}}
+    </:content>
+
+    <:overlay>
+      <PrimaryButton {{on "click" this.toggleSolution}} class="self-center">
```
rohitpaulk, line 43: Another minor naming note: 

```suggestion
      <PrimaryButton {{on "click" this.handleRevealSolutionButtonClick}} class="self-center">
```

I know that the current implementation is consistent with autofix-result.hbs, so no mistake on your part - that file seems to be a mistake though, we usually don't name click handlers like this. There are two problems here: 

1. We don't actually support "toggling", we only support unblurring/revealing a solution. This unintentionally suggests that toggling is possible. This adds a bunch of useless code - like needing to write `this.solutionIsBlurred = !this.solutionIsBlurred` instead of the simpler `this.solutionIsBlurred = false`

2. Our click handlers are are usually named like `this.handleClick` / `this.handleXYZClick`. This reduces the naming decisions we have to make + allows us to add functionality to the click handler without having to change the name. Hypothetical  example, if here we decided to show a popup before toggling the solution, we'd have to both change the implementation and also change the name to something like `openRevealSolutionConfirmationModal`. If the name was `handleRevealSolutionButtonClick`, we'd only need to change the implementation of the function, the name would still be valid.

app/controllers/course/stage/instructions.ts
```diff
@@ -54,13 +54,17 @@ export default class CourseStageInstructionsController extends Controller {
...
   get shouldShowLanguageGuide() {
-    return !this.model.courseStage.isFirst && !!this.languageGuide && this.featureFlags.canSeeLanguageGuidesForStage2;
+    return !this.model.courseStage.isFirst && !!this.languageGuide;
   }
 
   get shouldShowPrerequisites() {
     return !!this.prerequisiteInstructionsMarkdown;
   }
 
+  get shouldShowSolution() {
```
rohitpaulk, line 64: We're in "instructions.hbs" here, so this could apply to any stage. Since our feature flag is only for stage 2, maybe we could do this: 

```suggestion
  get shouldShowStage2Solution() {
```

app/controllers/course/stage/instructions.ts
```diff
@@ -54,13 +54,17 @@ export default class CourseStageInstructionsController extends Controller {
...
+    return this.featureFlags.canSeeSolutionsForStage2;
```
rohitpaulk, line 65:
```suggestion
    return this.featureFlags.canSeeSolutionsForStage2 && this.repository.secondStageSolution;
```

(Consistent with what we're doing in the other getters, and ensures we don't accidentally rely on the property in some place where a solution isn't available)

app/templates/course/stage/instructions.hbs
```diff
@@ -41,6 +41,7 @@
         @repository={{@model.activeRepository}}
         @courseStage={{@model.courseStage}}
         @shouldRecommendLanguageGuide={{this.shouldShowLanguageGuide}}
+        @shouldShowSolution={{this.shouldShowSolution}}
```
rohitpaulk, line 44:
```suggestion
        @shouldShowSolution={{this.shouldShowStage2Solution}}
```

(related to above comment)

The arg name is fine, since that's on `SecondStageInstructionsCard`.

rohitpaulk:
@andy1li no tests?

rohitpaulk (changes requested):
Notes added!

app/models/course-stage-solution.ts
```diff
@@ -1,8 +1,13 @@
 import Model, { attr, belongsTo } from '@ember-data/model';
+import { service } from '@ember/service';
+import ViewableMixin from 'codecrafters-frontend/mixins/viewable'; // eslint-disable-line ember/no-mixins
+import type AuthenticatorService from 'codecrafters-frontend/services/authenticator';
 import type CourseStageModel from 'codecrafters-frontend/models/course-stage';
 import type LanguageModel from 'codecrafters-frontend/models/language';
 
-export default class CourseStageSolutionModel extends Model {
+export default class CourseStageSolutionModel extends Model.extend(ViewableMixin) {
+  @service declare authenticator: AuthenticatorService;
```
rohitpaulk, line 9: @andy1li doesn't look like this is being used, is it?

rohitpaulk: Ah ignore, Viewable uses it

tests/acceptance/course-page/course-stage-solutions-test.js
```diff
@@ -0,0 +1,55 @@
...
+    await catalogPage.visit();
+    await catalogPage.clickOnCourse('Build your own Redis');
+
+    await coursePage.secondStageInstructionsCard.clickOnExpandStepButton();
+    await coursePage.secondStageInstructionsCard.clickOnCompleteStepButton();
+  });
+
+  test('can view solution', async function (assert) {
+    assert.ok(coursePage.secondStageInstructionsCard.hasRevealSolutionButton, 'Expect reveal solution button to be visible');
+    await percySnapshot('Course Stage Solutions - Before Reveal');
```
rohitpaulk, line 48:
```suggestion
    await percySnapshot('Second Stage Solution - Before Reveal');
```

(Bit more specific)

tests/acceptance/course-page/course-stage-solutions-test.js
```diff
@@ -0,0 +1,55 @@
...
+
+    await coursePage.secondStageInstructionsCard.clickOnRevealSolutionButton();
+    assert.false(coursePage.secondStageInstructionsCard.hasRevealSolutionButton, 'Expect reveal solution button to be hidden');
+    assert.ok(coursePage.secondStageInstructionsCard.hasFileDiffCard, 'Expect file diff card to be visible');
+    await percySnapshot('Course Stage Solutions - After Reveal');
```
rohitpaulk, line 53:
```suggestion
    await percySnapshot('Second Stage Solution - After Reveal');
```

tests/pages/components/course-page/course-stage-step/second-stage-instructions-card.ts
```diff
@@ -2,7 +2,12 @@ import { clickable, collection, isVisible, text } from 'ember-cli-page-object';
 
 export default {
   clickOnCompleteStepButton: clickable('[data-test-complete-step-button]'),
+  clickOnExpandStepButton: clickable('[data-test-expand-step-button]'),
+  clickOnRevealSolutionButton: clickable('[data-test-reveal-solution-button]'),
+
```
rohitpaulk, line 7: @andy1li let's not add newlines between statements like this - for reference, the rule we follow is: we only add newlines around statements that are multi-line (like `scrollIntoView` below). Everything else is just sorted alphabetically. More consistenty => Less cognitive load required + easier to automate refactoring / use AI for changes. 

The existing empty line between `scope` and `hasScreencastsLink` looks like a mistake - feel free to correct that.

rohitpaulk (approved):

Issue #1231, opened 2024-01-29
Use CodeMirror for rendering code/diffs


rohitpaulk:
Currently, our code rendering system is built in-house using Shiki for code highlighting. Some features we've added on top of this are: 

- Support for rendering diffs (green/red background for lines, +/- icons etc.)
- Support for expanding/collapsing unchanged diff sections
- Support for rendering comments (comment icons, expand/collapse comments etc.)

This is getting hard for us to extend, mostly because there isn't a solid foundation that is oriented around code editing. Let's transition away from this and use [CodeMirror](https://codemirror.net/) instead. 

# Rationale

We're already running into some limitations with the current system. 

## Current work

One example of a current limitation is the way we handle comments. 

Our comment buttons are currently anchored to the right of a line, so in most cases they're hidden by default. Even if you scroll to them, the disappear when you scroll back to read the comment: 

https://github.com/codecrafters-io/frontend/assets/3893573/fafb268e-9cd8-4471-ae9b-4f8a4ce47942

We want to them to be anchored to the right-most end of the editor container (not the line), like with Repl.it: 

https://github.com/codecrafters-io/frontend/assets/3893573/4e5aa506-ac28-4e9f-a7b0-36246f6aaebf

This is non-trivial to implement from scratch, since they need to be absolutely positioned wrt. the editor container in such a way that they're aware of a line's current Y-index. With CodeMirror, this'll be easier to do by using a system similar to their "gutter" feature (that renders icons on the left side, not right). You'll see that repl-it is using some kind of gutters plugin too. 

 ## Future work
 
 A lot of the features we want to build in the future will be easier to build if we're using CodeMirror. 
 
 - Select text and click on "explain this"
 - CodeIntelligence like GitHub, where you hover over a method and it shows you references, go-to defn. etc. 
 - Displaying line numbers along with diffs
 
 We might also add an in-browser editor at some point (thought it's unlikely that'll happen soon). 

# Scope

For starters, let's not add any new features but transition our existing features to use CodeMirror. I'm ordering these based on least used -> most used (we should ship PRs for the least used first, so we can iron out any issues early on before shipping)

## 1. Rendering buildpack code: [URL](https://app.codecrafters.io/courses/redis/admin/buildpacks/f9f43ed2-2147-4d62-8b79-bc0b0abd9814)

This only renders code, no diffs. No expand/collapse functionality, no comments functionality. 

[...]

PR #2400, opened 2024-11-17
Cleanup and unify CodeMirror Demo pages

VasylMarchuk:
Related to #1231

### Brief

In preparation for adding `FileDiffCardCM` component (CodeMirror-based clone of `FileDiffCard`) to the Demo Zone, this cleans up & unifies the Demo route pages.

### Details

This only involves the Demo page, nothing else. Changes to `FileContentsCard` Demo page are inspired by the upcoming addition of `FileDiffCardCM` to the Demo, and make all Demo pages behave similarly.

### More details

- Extracted all example Demo documents into `app/utils/code-mirror-documents`, to reuse them across Demo pages instead of cloning:
  - documents in the util are now generated from **diffs**
  - these diffs are much easier to work with than `document` & `originalDocument` pairs
  - example documents can now be used for diff-consuming components (such as `FileDiffCard`)
- Made all options from `FileContentsCard` Demo page persistent via query params
- Added `Reset all options` link to `FileContentsCard` Demo page
- Moved `DarkModeToggle` component to appear last in the Demo Zone navigation bar
- Extracted diff-parsing logic from `diff-to-document` helper into an util `parse-diff-as-document`, to reuse it in `code-mirror-documents` util

### Checklist

## Release Notes

- **New Features**
  - Introduced a "Reset all options" button in the file contents card for enhanced interactivity.
  - Added a streamlined management system for example documents using a predefined set instead of dynamic instantiation.

- **Improvements**
  - Enhanced handling of query parameters in the file contents card route for better URL management.
  - Simplified parsing logic in the document helper function for improved performance.

- **Bug Fixes**
  - Adjusted routing and display logic for links to ensure correct navigation between components.

- **Tests**
  - Added new test cases for the `parseDiffAsDocument` function to ensure reliability and error handling.

### DIFF
--- a/app/helpers/diff-to-document.ts
+++ b/app/helpers/diff-to-document.ts
@@ -1,30 +1,8 @@
 import { helper } from '@ember/component/helper';
-
-export function parseDiff(diff: string = '') {
-  const diffLines = diff.split('\n');
-
-  const current = [];
-  const original = [];
-
-  for (const line of diffLines) {
-    if (line.startsWith('-')) {
-      original.push(line.substring(1));
-    } else if (line.startsWith('+')) {
-      current.push(line.substring(1));
-    } else {
-      original.push(line.substring(1));
-      current.push(line.substring(1));
-    }
-  }
-
-  return {
-    current: current.join('\n'),
-    original: original.join('\n'),
-  };
-}
+import parseDiffAsDocument from 'codecrafters-frontend/utils/parse-diff-as-document';
 
 const diffToDocument = helper(function diffToDocument([diff = '']: [string | undefined]) {
-  return parseDiff(diff);
+  return parseDiffAsDocument(diff);
 });
 
 export default diffToDocument;
--- a/app/helpers/diff-to-document.ts
+++ b/app/helpers/diff-to-document.ts
@@ -1,6 +1,6 @@
 import { helper } from '@ember/component/helper';
 
-const diffToDocument = helper(function diffToDocument([diff = '']: [string | undefined]) {
+export function parseDiff(diff: string = '') {
   const diffLines = diff.split('\n');
 
   const current = [];
@@ -21,6 +21,10 @@ const diffToDocument = helper(function diffToDocument([diff = '']: [string | und
     current: current.join('\n'),
     original: original.join('\n'),
   };
+}
+
+const diffToDocument = helper(function diffToDocument([diff = '']: [string | undefined]) {
+  return parseDiff(diff);
 });
 
 export default diffToDocument;
--- a/app/router.ts
+++ b/app/router.ts
@@ -104,7 +104,7 @@ Router.map(function () {
 
   this.route('demo', function () {
     this.route('code-mirror');
-    this.route('dark-mode-toggle');
     this.route('file-contents-card');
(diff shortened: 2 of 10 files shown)


app/helpers/diff-to-document.ts
```diff
@@ -1,6 +1,6 @@
 import { helper } from '@ember/component/helper';
 
-const diffToDocument = helper(function diffToDocument([diff = '']: [string | undefined]) {
+export function parseDiff(diff: string = '') {
```
rohitpaulk, line 3: @VasylMarchuk is there a downside to keeping this as it was? I think that's what we do in other helpers. Not a big deal but if we can stay consistent might as well.

app/router.ts
```diff
@@ -104,7 +104,7 @@ Router.map(function () {
 
   this.route('demo', function () {
     this.route('code-mirror');
-    this.route('dark-mode-toggle');
     this.route('file-contents-card');
+    this.route('dark-mode-toggle');
```
rohitpaulk, line 108: These are usually alphabetically sorted, so the previous version was correct

rohitpaulk: (Anywhere order doesn't matter, we try to stick to alphabetical ordering)

rohitpaulk (approved):

app/helpers/diff-to-document.ts
VasylMarchuk: I want to reuse the parsing method that the helper uses in a different place (not the helper itself), maybe I can move it into a separate util?

rohitpaulk: Ah, gotcha - yep let's move that to lib then! Makes it more obvious that it will be reused

VasylMarchuk: Done, moved into `utils/parse-diff-as-document` + added unit tests (copied from helper integration tests, but in unit syntax)

app/utils/parse-diff-as-document.ts
```diff
@@ -0,0 +1,22 @@
...
+  for (const line of diffLines) {
+    if (line.startsWith('-')) {
+      original.push(line.substring(1));
+    } else if (line.startsWith('+')) {
+      current.push(line.substring(1));
+    } else {
+      original.push(line.substring(1));
+      current.push(line.substring(1));
+    }
+  }
```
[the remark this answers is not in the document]
VasylMarchuk: Good idea for a separate PR if this comes up as an issue

app/utils/parse-diff-as-document.ts
```diff
@@ -0,0 +1,22 @@
...
+
+  return {
+    current: current.join('\n'),
+    original: original.join('\n'),
+  };
+}
```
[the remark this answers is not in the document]
VasylMarchuk: Our diffs are not exactly classical diffs, they don't contain headers or hunks

app/utils/parse-diff-as-document.ts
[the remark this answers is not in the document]
VasylMarchuk: Considering that our diffs are not really classical diffs, we'll handle this if we ever want to migrate to true diffs.

PR #2500, opened 2024-12-26
Add temporary card "I'm working on improving our stage 2 experience"

andy1li:
image

**Checklist**:

## Summary by CodeRabbit

- **New Features**
	- Introduced a temporary interactive card featuring a character named Andy for user feedback during the second stage of the course.
	- The card is conditionally displayed based on the course stage and slug.

### DIFF
--- a/app/templates/course/stage/instructions.hbs
+++ b/app/templates/course/stage/instructions.hbs
@@ -84,16 +84,9 @@
       </div>
     {{/if}}
 
-    {{! TODO: Bring this back? }}
-    {{!-- {{#if this.shouldShowPrerequisites}}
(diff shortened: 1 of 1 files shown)


app/templates/course/stage/instructions.hbs
```diff
@@ -91,6 +91,33 @@
 
     <CoursePage::CourseStageStep::YourTaskCard @repository={{@model.activeRepository}} @courseStage={{@model.courseStage}} />
 
+    {{! TODO: Remove this temporary card  }}
+    {{#if (and @model.courseStage.isSecond (eq @model.courseStage.course.slug "shell"))}}
```
rohitpaulk, line 95: @andy1li thoughts on keeping this a bit higher up? Worried that not enough people will see this since it falls below the your task card.

rohitpaulk (approved):
Looks good but think it'd be better to place a bit higher up where there's more visibility

PR #2588, opened 2025-01-27
Refactor billing status display and add new component

ryan-gang:
This pull request includes several commits that refactor the billing status display in the codebase. The `BillingStatusDisplayService` has been removed as it was no longer used. The rendering of the billing status badge has been simplified by introducing a new component called `BillingStatusDisplay::BillingStatusBadge`. Additionally, a new component called `BillingStatusDisplay::DiscountTimerBadge` has been added to handle promotional discounts. These changes improve the code structure and make it easier to handle different billing status scenarios.

**Checklist**:

## Summary by CodeRabbit

- **Refactor**
	- Simplified billing status display logic across multiple components
	- Consolidated billing-related badges into a single `BillingStatusBadge` component
	- Removed `BillingStatusDisplayService`

- **New Features**
	- Added size configuration for various billing-related badges
	- Introduced dynamic badge rendering based on user status

- **Chores**
	- Updated component and service registrations
	- Streamlined UI element visibility logic


app/components/billing-status-display/billing-status-badge.hbs
```diff
@@ -0,0 +1,25 @@
+<div ...attributes>
+  {{#if this.shouldShowVipBadge}}
+    <Header::VipBadge @size={{@size}} />
+  {{else if this.shouldShowMemberBadge}}
+    <Header::MemberBadge @size={{@size}} />
+  {{else if this.activeDiscountForYearlyPlan}}
```
rohitpaulk, line 6:
```suggestion
  {{else if (and this.activeDiscountForYearlyPlan this.canSeeDiscountCoundown)}}
```

(This way, if the user can't see the discount countdown, the next condition will trigger - else we won't end up rendering the upgrade button for many users)

app/components/billing-status-display/billing-status-badge.hbs
```diff
@@ -0,0 +1,25 @@
...
+    {{#if this.canSeeDiscountCountdown}}
+      <BillingStatusDisplay::DiscountTimerBadge @discount={{this.activeDiscountForYearlyPlan}} @size={{@size}} />
+    {{/if}}
+  {{else if this.shouldShowFreeWeeksLeftButton}}
+    <Header::FreeWeeksLeftButton @size={{@size}} />
+  {{else if this.shouldShowUpgradeButton}}
+    <PrimaryLinkButton
+      @size={{this.linkButtonSize}}
+      @route="pay"
+      class="{{if (eq this.linkButtonSize 'extra-small') 'mr-3'}}"
```
rohitpaulk, line 16: @ryan-gang why do we have to specify `mr-3` manually here? Shouldn't the `...attributes` in the parent cover this?

app/components/billing-status-display/discount-timer-badge.hbs
```diff
@@ -1,16 +1,16 @@
 <LinkTo
   @route="pay"
   class="flex bg-gradient-to-b from-gray-600 to-gray-900 border border-yellow-600 rounded shadow-sm
-    {{if (eq @size 'small') 'px-1.5 py-1'}}
-    {{if (eq @size 'large') 'px-2 py-1.5'}}"
+    {{if (eq @size 'small') 'px-1.5 py-1 mr-3 text-xs'}}
```
rohitpaulk, line 4: Doesn't make sense that we've have to set `mr-3` here - isn't this something that the parent can pass in?

app/components/billing-status-display/discount-timer-badge.ts
```diff
@@ -38,6 +38,6 @@ export default class DiscountTimerBadgeComponent extends Component<Signature> {
 
 declare module '@glint/environment-ember-loose/registry' {
   export default interface Registry {
-    'BillingStatusBadge::DiscountTimerBadge': typeof DiscountTimerBadgeComponent;
+    'BillingStatusDisplay::DiscountTimerBadge': typeof DiscountTimerBadgeComponent;
```
rohitpaulk, line 41:
```suggestion
    'BillingStatusBadge::DiscountTimerBadge': typeof DiscountTimerBadgeComponent;
```

When we have nested components like this, we usually name `BillingStatusBadge` as one component, and `BillingStatusBadge::WhateverBadge` as nested components. This avoids the extra name for the directory (`BillingStatusDisplay` in this case).

app/components/billing-status-display/billing-status-badge.ts
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  get shouldShowVipBadge(): boolean {
+    return !!this.currentUser && this.currentUser.isVip;
+  }
+}
+
+declare module '@glint/environment-ember-loose/registry' {
+  export default interface Registry {
+    'BillingStatusDisplay::BillingStatusBadge': typeof BillingStatusDisplayComponent;
```
rohitpaulk, line 67:
```suggestion
    BillingStatusBadge: typeof BillingStatusDisplayComponent;
```

(Notes below)

app/components/header/free-weeks-left-button.hbs
```diff
@@ -1,4 +1,9 @@
-<PrimaryLinkButton @size="extra-small" @route="pay" class="mr-4" data-test-free-weeks-left-button>
+<PrimaryLinkButton
+  @size={{this.linkButtonSize}}
+  @route="pay"
+  class="{{if (eq this.linkButtonSize 'extra-small') 'mr-3'}}"
```
rohitpaulk, line 4: Same note as elsewhere - we shouldn't have to set margins inside components and should instead have the parent pass them in. 

The principle to follow here is: a component is roughly a "box", and it should ideally only control everything inside the box, and nothing outside the box. Margins affect outer layout, which is the responsibility of the parent component / template.

app/components/header/member-badge.hbs
```diff
@@ -1,4 +1,11 @@
-<PrimaryLinkButton @size="extra-small" @route="membership" data-test-member-badge ...attributes>
+<PrimaryLinkButton
+  @size={{this.linkButtonSize}}
+  @route="membership"
+  class="{{if (eq this.linkButtonSize 'extra-small') 'mr-3'}}"
```
rohitpaulk, line 4: Same note as above

app/components/header/vip-badge.hbs
```diff
@@ -1,4 +1,10 @@
-<div class="text-xs text-white font-semibold bg-teal-500 rounded px-1.5 py-1 flex items-center" data-test-vip-badge ...attributes>
+<div
+  class="text-white font-semibold bg-teal-500 rounded flex items-center justify-center
+    {{if (eq this.size 'small') 'text-xs px-2 py-1 mr-3'}}
```
rohitpaulk, line 3: Same note as above re: margins

app/components/header/vip-badge.ts
```diff
@@ -5,6 +5,9 @@ import type AuthenticatorService from 'codecrafters-frontend/services/authentica
 
 interface Signature {
   Element: HTMLDivElement;
+  Args: {
+    size: 'small' | 'large';
+  };
 }
 
 export default class VipBadgeComponent extends Component<Signature> {
```
rohitpaulk, line 13: @ryan-gang now that we have a parent component for this, `MemberBadge` and `VipBadge` can both be moved into `BillingStatusBadge::MemberBadge` and `BillingStatusBadge::VipBadge`. 

Earlier the naming made sense since they were only used in `Header` (which is different from `CoursePageHeader`). Now that they're used in two places, `BillingStatusBadge` seems like a better home.

app/components/header.hbs
```diff
@@ -58,25 +58,7 @@
...
-            <Header::FreeWeeksLeftButton class="mr-3" />
-          {{else if this.billingStatusDisplay.shouldShowUpgradeButton}}
-            <PrimaryLinkButton @size="extra-small" @route="pay" class="mr-4" data-test-upgrade-button>
-              <span class="flex items-center gap-x-1">
-                <span>Upgrade</span>
-                {{svg-jar "lock-open" class="w-4"}}
-              </span>
-            </PrimaryLinkButton>
-          {{/if}}
+          <BillingStatusDisplay::BillingStatusBadge @size="small" />
```
rohitpaulk, line 61:
```suggestion
          <BillingStatusDisplay::BillingStatusBadge @size="small" class="mr-3" />
```

(This is the correct place to handle margins)

The `mr-4` that was previously applied to the Upgrade button above seems like a mistake, I think it's okay to remain consistent and use `mr-3` for all badges.

ryan-gang:
Also moved the `FreeWeeksLeftButton` component into the `billing-status-badge` dir.

ryan-gang:
Ready to deploy.

PR #2633, opened 2025-02-05
Refactor concept engagement block counting and progress display

ryan-gang:
Improve the logic for counting total and completed blocks in the `ConceptEngagementModel`. Update the concept progress component to display counts instead of percentages, enhancing clarity and conciseness.

**Checklist**:

- [x] I've thoroughly self-reviewed may changes

## Summary by CodeRabbit

- **New Features**
  - The progress indicator now displays completed blocks out of the total blocks, providing a clearer overview of engagement.
  - Introduced a new `BlockGroup` interface to enhance block processing.

- **Refactor**
  - Streamlined the progress display by removing the percentage and remaining blocks indicators for a simplified view.
  - Centralized block group management within the `ConceptModel`.

- **Tests**
  - Updated test cases to reflect the new progress display format, focusing on completed blocks instead of percentages or remaining blocks.

### DIFF
--- a/app/components/concept-progress.hbs
+++ b/app/components/concept-progress.hbs
@@ -6,9 +6,9 @@
 
     <div class="flex-shrink-0 ml-3 text-gray-600 text-sm font-medium">
       <AnimatedContainer>
-        {{#animated-value @latestConceptEngagement.currentProgressPercentage rules=this.rules duration=200 as |v|}}
+        {{#animated-value @latestConceptEngagement.completedBlocksCount rules=this.rules duration=200 as |c|}}
           <div>
-            {{v}}%
+            {{c}}/{{@latestConceptEngagement.totalBlocksCount}}
           </div>
         {{/animated-value}}
       </AnimatedContainer>
--- a/app/components/concept-progress.hbs
+++ b/app/components/concept-progress.hbs
@@ -6,9 +6,9 @@
 
     <div class="flex-shrink-0 ml-3 text-gray-600 text-sm font-medium">
       <AnimatedContainer>
-        {{#animated-value @latestConceptEngagement.completedBlocksCount rules=this.rules duration=200 as |c|}}
+        {{#animated-value @latestConceptEngagement.completedBlockGroupsCount rules=this.rules duration=200 as |c|}}
           <div>
-            {{c}}/{{@latestConceptEngagement.totalBlocksCount}}
+            {{c}}/{{@latestConceptEngagement.totalBlockGroupsCount}}
           </div>
         {{/animated-value}}
       </AnimatedContainer>
--- a/app/components/concept-progress.hbs
+++ b/app/components/concept-progress.hbs
@@ -14,17 +14,4 @@
       </AnimatedContainer>
     </div>
   </div>
-
-  {{#if (lt @latestConceptEngagement.currentProgressPercentage 100)}}
-    <div class="flex-shrink-0 mt-2 text-gray-600 text-xs">
-      <AnimatedContainer>
-        {{#animated-value @latestConceptEngagement.remainingBlocksCount rules=this.rules duration=200 as |v|}}
-          <div>
-            <span class="font-medium">{{v}}</span>
-            blocks left
-          </div>
-        {{/animated-value}}
-      </AnimatedContainer>
-    </div>
-  {{/if}}
 </div>
--- a/app/components/concept/index.ts
+++ b/app/components/concept/index.ts
@@ -118,11 +118,10 @@ export default class ConceptComponent extends Component<Signature> {
 
   @action
   async handleContinueButtonClick() {
-    if (this.currentBlockGroupIndex === this.allBlockGroups.length - 1) {
+    this.updateLastRevealedBlockGroupIndex(this.currentBlockGroupIndex + 1);
+    this.enqueueConceptEngagementUpdate.perform();
+    if (this.currentBlockGroupIndex === this.allBlockGroups.length) {
       this.hasFinished = true;
-    } else {
-      this.updateLastRevealedBlockGroupIndex(this.currentBlockGroupIndex + 1);
-      this.enqueueConceptEngagementUpdate.perform();
     }
 
     this.analyticsEventTracker.track('progressed_through_concept', {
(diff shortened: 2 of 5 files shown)


rohitpaulk:
- Completing a question should always trigger a change
- We shouldn't have +2 jumps, only +1

Try here: https://frontend-52e5p7jee.ccio.dev/concepts/network-protocols

app/components/concept-progress.hbs
```diff
@@ -6,25 +6,12 @@
 
     <div class="flex-shrink-0 ml-3 text-gray-600 text-sm font-medium">
       <AnimatedContainer>
-        {{#animated-value @latestConceptEngagement.currentProgressPercentage rules=this.rules duration=200 as |v|}}
+        {{#animated-value @latestConceptEngagement.completedBlocksCount rules=this.rules duration=200 as |c|}}
```
rohitpaulk, line 9:
```suggestion
        {{#animated-value @latestConceptEngagement.completedBlockGroupsCount rules=this.rules duration=200 as |c|}}
```

app/components/concept-progress.hbs
```diff
@@ -6,25 +6,12 @@
...
           <div>
-            {{v}}%
+            {{c}}/{{@latestConceptEngagement.totalBlocksCount}}
```
rohitpaulk, line 11:
```suggestion
            {{c}}/{{@latestConceptEngagement.totalBlockGroupsCount}}
```

app/components/concept/index.ts
```diff
@@ -22,7 +22,7 @@ interface Signature {
   Element: HTMLDivElement;
 }
 
-interface BlockGroup {
+export interface BlockGroup {
```
rohitpaulk, line 25: Let's move this out into `models/concept.ts` - so that models don't need to import from the view layer

app/models/concept-engagement.ts
```diff
@@ -10,10 +12,28 @@ export default class ConceptEngagementModel extends Model {
   @attr('date') declare lastActivityAt: Date;
   @attr('date') declare startedAt: Date;
 
-  get remainingBlocksCount() {
-    const allBlocks = this.concept.parsedBlocks;
-    const completedBlocksCount = Math.round((this.currentProgressPercentage / 100) * allBlocks.length);
+  @cached
+  get allBlockGroups(): BlockGroup[] {
```
rohitpaulk, line 16: Duplication, move into `concept.blockGroups` instead

app/models/concept-engagement.ts
```diff
@@ -10,10 +12,28 @@ export default class ConceptEngagementModel extends Model {
...
+
+      if (block.isInteractable || groups.length === 0) {
+        groups.push({ index: groups.length, blocks: [] });
+      }
+
+      return groups;
+    }, [] as BlockGroup[]);
+  }
+
+  get completedBlocksCount() {
```
rohitpaulk, line 32:
```suggestion
  get completedBlockGroupsCount() {
```

rohitpaulk (changes requested):
Let's take another look after changes are applied

app/models/concept-engagement.ts
```diff
@@ -10,10 +10,33 @@ export default class ConceptEngagementModel extends Model {
...
   @attr('date') declare startedAt: Date;
 
-  get remainingBlocksCount() {
-    const allBlocks = this.concept.parsedBlocks;
-    const completedBlocksCount = Math.round((this.currentProgressPercentage / 100) * allBlocks.length);
+  get completedBlockGroupsCount() {
+    const completedBlocksIndex = Math.round((this.currentProgressPercentage / 100) * this.totalBlocksCount);
 
-    return allBlocks.length - completedBlocksCount;
+    return this.parseIntoBlockGroups(this.concept.parsedBlocks.slice(0, completedBlocksIndex)).length;
```
rohitpaulk, line 16: Can't we move all the block group parsing stuff into `Concept#blockGroups`? And here, we could just multiply `currentProgressPercentage` by `Concept#blockGroups.length` to get the `completedBlockGroupsCount`? Not sure why we have to fetch `completedBlockIndex` when we're only dealing with block groups and not blocks at this layer.

rohitpaulk (changes requested):
Added q

app/models/concept.ts
```diff
@@ -30,6 +36,25 @@ export default class ConceptModel extends Model {
   @equal('status', 'draft') declare statusIsDraft: boolean;
   @equal('status', 'published') declare statusIsPublished: boolean;
 
+  @cached
+  get allBlockGroups(): BlockGroup[] {
```
rohitpaulk, line 40:
```suggestion
  get blockGroups(): BlockGroup[] {
```

(We don't need "all" here since there's no concept of all/completed - made sense within concept engagement but not here)

app/components/concept/index.ts
```diff
@@ -135,11 +118,11 @@ export default class ConceptComponent extends Component<Signature> {
 
   @action
   async handleContinueButtonClick() {
-    if (this.currentBlockGroupIndex === this.allBlockGroups.length - 1) {
+    this.updateLastRevealedBlockGroupIndex(this.currentBlockGroupIndex + 1);
+    this.enqueueConceptEngagementUpdate.perform();
+
+    if (this.currentBlockGroupIndex === this.allBlockGroups.length) {
```
rohitpaulk, line 124: @ryan-gang this seems invalid - `currentBlockGroupIndex` shouldn't go out of bounds of `this.allBlockGroups` no?

app/models/concept-engagement.ts
```diff
@@ -10,10 +10,31 @@ export default class ConceptEngagementModel extends Model {
...
+
+  get totalBlockGroupsCount() {
+    return this.concept.allBlockGroups.length;
+  }
+
+  get totalBlocksCount() {
+    return this.concept.parsedBlocks.length;
+  }
+
+  convertBlockProgressIntoBlockGroupProgress(allBlockGroups: BlockGroup[], completedBlocksCount: number) {
```
rohitpaulk, line 27: This function isn't re-used anywhere, so the logic can remain within `completedBlockGroupsCount` directly. 

Just add `completedBlocksCount` as one method, and then use that within `completedBlockGroupsCount` using the specified logic.

rohitpaulk (changes requested):
Seems incorrect that "lastRevealedBlockGroupIndex" would go out of bounds and point to a block group that doesn't exist - let's revisit this aspect please

rohitpaulk (approved):

PR #2651, opened 2025-02-12
Refactor and enhance question card option with confetti animation

ryan-gang:
Convert the question card option component to TypeScript, add a confetti animation for correct answers, and improve the component's functionality by persisting the confetti state and calculating the confetti origin based on mouse position or element center.

**Checklist**:

### DIFF
--- a/app/components/concept-page/question-card-option.hbs
+++ b/app/components/concept-page/question-card-option.hbs
@@ -7,6 +7,7 @@
     {{if this.isSelectedAndUnsubmitted 'bg-gray-100'}}
     {{if this.isSelectedAndCorrect 'bg-green-200'}}
     "
+  {{did-update this.fireCorrectAnswerConfetti @isSubmitted}}
   ...attributes
 >
   <div
--- a/app/components/concept-page/question-card-option.hbs
+++ b/app/components/concept-page/question-card-option.hbs
@@ -5,9 +5,8 @@
     {{if this.isUnselected 'hover:bg-gray-100 focus:bg-gray-100'}}
     {{if this.isUnselectedAndCorrect 'bg-green-200'}}
     {{if this.isSelectedAndUnsubmitted 'bg-gray-100'}}
-    {{if this.isSelectedAndCorrect 'bg-green-200'}}
-    "
-  {{did-update this.fireCorrectAnswerConfetti @isSubmitted}}
+    {{if this.isSelectedAndCorrect 'bg-green-200'}}"
+  {{did-insert this.fireCorrectAnswerConfetti}}
   ...attributes
 >
   <div
--- a/app/components/concept-page/question-card-option.ts
+++ b/app/components/concept-page/question-card-option.ts
@@ -1,6 +1,19 @@
 import Component from '@glimmer/component';
 
-export default class QuestionCardOptionComponent extends Component {
+interface QuestionOption {
+  is_correct: boolean;
+  isSelected: boolean;
+}
+
+interface QuestionCardOptionArgs {
+  option: QuestionOption;
+  isSubmitted: boolean;
+}
+
+export default class QuestionCardOptionComponent extends Component<QuestionCardOptionArgs> {
+  @service declare confetti: ConfettiService;
+  @tracked hasShownConfetti = false;
+
   get isCorrect() {
     return this.args.option.is_correct;
   }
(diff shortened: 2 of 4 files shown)


app/components/concept-page/question-card-option.hbs
```diff
@@ -5,8 +5,8 @@
     {{if this.isUnselected 'hover:bg-gray-100 focus:bg-gray-100'}}
     {{if this.isUnselectedAndCorrect 'bg-green-200'}}
     {{if this.isSelectedAndUnsubmitted 'bg-gray-100'}}
-    {{if this.isSelectedAndCorrect 'bg-green-200'}}
-    "
+    {{if this.isSelectedAndCorrect 'bg-green-200'}}"
+  {{did-insert this.fireCorrectAnswerConfetti}}
```
rohitpaulk, line 9: Let's fire confetti from the "Success" message, and not from the option itself

app/components/concept-page/question-card.ts
```diff
@@ -16,11 +16,16 @@ interface Signature {
 
 export default class QuestionCardComponent extends Component<Signature> {
   @tracked selectedOptionIndex: number | null = null;
+  @tracked _hasShownConfetti = false;
```
rohitpaulk, line 19:
```suggestion
  @tracked hasShownConfetti = false;
```

We don't need a separate underscored version

rohitpaulk (dismissed):
Notes added

app/components/concept-page/question-card-option.ts
```diff
@@ -29,3 +46,9 @@ export default class QuestionCardOptionComponent extends Component {
     return this.args.isSubmitted && !this.args.option.isSelected && this.args.option.is_correct;
   }
 }
```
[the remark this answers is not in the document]
ryan-gang: The confetti functionality is moved to the `questionCard` component.
Where it is injected, tracked and fired.

app/components/concept-page/question-card-option.ts
```diff
@@ -1,6 +1,20 @@
 import Component from '@glimmer/component';
 
-export default class QuestionCardOptionComponent extends Component {
+export interface Signature {
+  Element: HTMLButtonElement;
+
+  Args: {
+    option: {
+      is_correct: boolean;
+      isSelected: boolean;
```
rohitpaulk, line 9: There's a mixture of camelCase and snake_case here. Any way we can avoid this?

rohitpaulk: At the very least let's add a comment here - I know that this is old code, but we didn't intend to have this kind of mixture of casing

rohitpaulk: Also, let's try to import `Option` from `ConceptQuestion` model rather than re-defining it here?

app/components/concept-page/question-card.hbs
```diff
@@ -43,6 +42,7 @@
       <div
         data-test-question-submitted
         class="p-3 sm:p-4 shadow-sm border rounded {{if this.selectedOptionIsCorrect 'border-green-400 bg-green-50' 'border-gray-300 bg-white'}}"
+        {{did-insert this.fireCorrectAnswerConfetti}}
```
rohitpaulk, line 45:
```suggestion
        {{did-insert this.handleDidInsertCorrectAnswerEmoji}}
```

1. We usually don't give actions names since over time the actions will evolve to do multiple stuff, and it's easier to just name them by how they're triggered
2. Let's try adding this below, to the 🎉 emoji?

app/components/concept-page/question-card.ts
```diff
@@ -15,6 +17,9 @@ interface Signature {
 }
 
 export default class QuestionCardComponent extends Component<Signature> {
+  @service declare confetti: ConfettiService;
+
+  @tracked hasShownConfetti = false;
```
rohitpaulk, line 22:
```suggestion
  @tracked hasFiredConfetti = false;
```

(Minor change - makes it a bit more obvious what the action was)

rohitpaulk (approved):

app/components/concept-page/question-card.hbs
rohitpaulk: The reason to attach to the emoji is that it's on the left, and that's where the user needs to start reading from. Now it fires in the center

Issue #2607, opened 2025-01-31
Statically rendered user pages + concept pages


rohitpaulk:
We've been pre-rendering track pages and challenge pages by enumerating over all available tracks and challenges. We now need to find a way to statically generate user pages (there are 1000s of these), and concept pages (we'll have 100s of these in the near future). 

Vercel's [Prerender functions](https://vercel.com/docs/build-output-api/v3/primitives#prerender-functions) API is supposed to allow for generating these on demand and caching. Let's implement this and also see if we can expose this as a generic ember plugin that others can use.

rohitpaulk:
@VasylMarchuk can you look into this please?

rohitpaulk:
@VasylMarchuk this is a bit time-sensitive, let's try to have it out by Feb 10

rohitpaulk:
For additional context - the reason we can't use the existing pre-rendering setup for these is that there are 1000s of these pages, and our deploys would take hours if we were to pre-render all of these pages.

VasylMarchuk:
Looking into this. Also, is my suspicion correct that since we're now going to render user-generated content (their user pages), we need to fix the DOMPurify `sanitize` issue? @rohitpaulk 

<img width="679" alt="Image" src="https://github.com/user-attachments/assets/8a5e4596-aaed-4dad-bf71-a759ee4158f2" />

rohitpaulk:
@VasylMarchuk yeah, we'll need to do something about that since user pages can contain UGC. It's okay if we do this in a hacky manner for now though - for example, if we just HTML-escaped all content when in fastboot mode (so it doesn't insert things like <script> tags), that might work. We can then re-visit getting DOMPurify to work properly in a different issue.

rohitpaulk:
This is done now! All's that pending is to make the 404 for concept pages work.

PR #2653, opened 2025-02-12
Prerender User Profile pages with Vercel Prerender Functions

VasylMarchuk:
First part of #2607 

### Brief

This adds server-side rendering for `/users/:username` route, based on Vercel's [Prerender Functions](https://vercel.com/docs/build-output-api/v3/primitives#prerender-functions).

### Details

- Added a correct page title for `/users/user` template
- Set proper OG meta tags for the `user` route
- Used `peekRecord` in `user` route model hook to fix blinking after using FastBoot-rendered html
- Cleaned up `vercel.json` file, put _redirects_ above _rewrites_ to match their order on Vercel
- Added Vercel build script in `/scripts/build-vercel-output.sh`:
  - it runs `npm run build` and checks the output for errors
  - then it properly installs Vercel functions:
    - copies all functions from project's `/vercel-functions` into `/vercel/output/functions`
    - copies the built `dist` folder into each function
    - runs `npm install` inside of each function
  - much simpler `buildCommand` and `installCommand` in `vercel.json`
- Slightly improved NPM `build` task with a better break-down of steps (added `build:ember` and `build:versions`)
- Added new `/prerender/_prerender-ember-route` Vercel prerender function
  - `maxDuration` is set to `15 seconds`
  - `memory` is set to `3009 MB`, which automatically [gives the function better VCPU](https://vercel.com/docs/functions/configuring-functions/memory#memory-/-cpu-type)
- Added an alias function `/prerender/users/[user]` which reuses `/prerender/_prerender-ember-route` function
- Made `/prerender/users/[user]` function handle `/users/:user` route instead of the middleware
  - `expiration` time is currently set to `24 hours`
- Removed `/users/:username` route handling from the middleware and cleaned it up

### Checklist

## Summary by CodeRabbit

- **New Features**
  - Enhanced user profile pages now display personalized titles.
  - Improved metadata handling and prerendering deliver faster page loads and better SEO on user profiles, contests, and concept pages.
  - Updated social sharing meta tags ensure accurate previews for contest and concept routes.
  - New functionality for prerendering user-specific pages with dynamic query parameter support.

- **Chores**
  - Streamlined build and deployment processes for more reliable releases.
  - Revamped routing and redirect configurations improve overall navigation and page delivery.
  - Introduced a new Bash script to automate the Vercel build process.
  - Updated installation behavior to suppress funding and audit messages during npm install.


VasylMarchuk:
@coderabbitai review

vercel-functions/prerender-user-profile-page.func/index.js
```diff
@@ -0,0 +1,55 @@
...
+
+    // Customize the sandbox globals
+    buildSandboxGlobals(defaultGlobals) {
+      return Object.assign({}, defaultGlobals, {
+        AbortController,
+      });
+    },
+
+    maxSandboxQueueSize: 1,
+  });
```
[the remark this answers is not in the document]
VasylMarchuk: I believe un such a case we want the serverless function to actually crash, instead of showing a 404 page

vercel-functions/prerender-user-profile-page.func/index.js
```diff
@@ -0,0 +1,55 @@
...
+export default async function (request, response) {
+  const { username } = request.query;
+
+  // Check if username query parameter is provided
+  if (!username) {
+    console.error('Missing "username" query parameter');
+    response.redirect('/404');
+
+    return;
+  }
```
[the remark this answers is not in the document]
VasylMarchuk: I don't think we need to do validation here, wrong usernames will be handled by ember routing and will result in a redirect response from FastBoot anyway

vercel-functions/prerender-user-profile-page.func/index.js
```diff
@@ -0,0 +1,55 @@
...
+  const result = await app.visit(`/users/${username}`);
+  const statusCode = result._fastbootInfo.response.statusCode;
+
+  // Redirect to 404 page if the status code is not 200
+  if (statusCode !== 200) {
+    console.warn('Error parsing FastBoot response, statusCode was:', statusCode);
+    response.redirect('/404');
+
+    return;
+  }
```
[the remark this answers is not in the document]
VasylMarchuk: Interesting, but not sure what caching headers we want exactly, and for how long, will look into this

package.json
```diff
@@ -13,7 +13,11 @@
   "scripts": {
     "tailwind:build": "npx tailwindcss -i ./app/tailwind-input.css -o ./app/styles/tailwind-compiled.css",
     "tailwind:watch": "npx tailwindcss -i ./app/tailwind-input.css -o ./app/styles/tailwind-compiled.css --watch",
-    "build": "npm --version && npm run tailwind:build && ember build --environment=production",
+    "build": "npm run build:versions && npm run tailwind:build && npm run build:ember",
+    "build:versions": "echo \"npm: $(npm --version)\" && ember --version",
+    "build:ember": "ember build --environment=production",
+    "build:vercel:dist": "./scripts/build-vercel-dist.sh",
```
rohitpaulk, line 19: Overall this idea of separate build scripts is pretty confusing - would see if we can stick to a single `npm run build`. I don't think a dev is going to invoke any of these independently.

There's also a bit of a weird dependency here - `build:vercel:dist` invokes a script that in turn invokes `npm run build`. It's not circular per-se, but still pretty weird. 

Can we get rid of `build:vercel:dist` and `build:vercel:functions`, and just rely on the scripts directly?

scripts/build-vercel-dist.sh
```diff
@@ -0,0 +1,21 @@
+#! /bin/bash
```
rohitpaulk, line 1: @VasylMarchuk let's just merge these into a single script (`scripts/build-vercel-output.sh`)? I don't see a place where they're invoked separately.

scripts/build-vercel-dist.sh
```diff
@@ -0,0 +1,21 @@
...
+
+LOG_FILENAME="build-output.log"
+ERROR_STRING="error"
```
rohitpaulk, line 4: @VasylMarchuk would refrain from indirection unless we're going to re-use a variable, or it's a very "configurable" kind of value. In this case I'd just inline the value, extracting it out doesn't make it any easier to understand what's happening here.

Extracting `LOG_FILENAME` is fine since it's re-used multiple times.

scripts/build-vercel-functions.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+
+VERCEL_FUNCTIONS_SOURCE="vercel-functions"
```
rohitpaulk, line 3: Same note as above -- try to avoid indirection unless it truly helps with readability. Here I don't think it applies to most values extracted out as constants

scripts/build-vercel-functions.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
```
rohitpaulk, line 1: We could use `set -e` and/or `set -o pipefail`  here, that'd prevent having to add `|| exit 1` everywhere.

vercel-functions/prerender-user-profile-page.func/index.js
```diff
@@ -0,0 +1,55 @@
...
+
+  // Get the HTML content of the FastBoot response
+  const html = await result['html'](); // Weird VSCode syntax highlighting issue if written as result.html()
+
+  // Define meta tag values
+  const pageTitle = `${username}'s CodeCrafters Profile`;
+  const pageImageUrl = `https://og.codecrafters.io/api/user_profile/${username}`; // TODO: Read `metaTagUserProfilePictureBaseURL` from page config
```
rohitpaulk, line 47: Let's remove the TODO, and remove the config property too - not worth the indirection

vercel-functions/prerender-user-profile-page.func/index.js
```diff
@@ -0,0 +1,55 @@
...
+    response.redirect('/404');
+
+    return;
+  }
+
+  // Get the HTML content of the FastBoot response
+  const html = await result['html'](); // Weird VSCode syntax highlighting issue if written as result.html()
+
+  // Define meta tag values
+  const pageTitle = `${username}'s CodeCrafters Profile`;
```
rohitpaulk, line 46: Hmm, kinda weird that we need to do this separately. Can we not invoke the code that would naturally end up triggering for this? I guess that's `middleware.js`?

vercel.json
```diff
@@ -1,33 +1,21 @@
 {
-  "buildCommand": "npm run build  > >(tee -a build-output.log) 2> >(tee -a build-output.log >&2) && ! (cat build-output.log | grep -qi error) && echo \"No errors found in build output\"",
+  "buildCommand": "npm run build:vercel:dist && npm run build:vercel:functions",
```
rohitpaulk, line 2: Unless there's a good reason to separate these out, I'd keep just a single script

vercel-functions/prerender-user-profile-page.func/index.js
```diff
@@ -0,0 +1,55 @@
...
+      return Object.assign({}, defaultGlobals, {
+        AbortController,
+      });
+    },
+
+    maxSandboxQueueSize: 1,
+  });
+
+  // Visit the user profile page
+  const result = await app.visit(`/users/${username}`);
```
rohitpaulk, line 31: It sucks that we have to write this individually for each route - @VasylMarchuk wonder if there's a way to make this work for "any" route? 

For example, maybe we could use a single `/prerender.func` as a serverless function, and have that take `?path=/users/abcd` as a query param?

rohitpaulk (changes requested):
Added one round of comments!

vercel-functions/prerender-user-profile-page.func/index.js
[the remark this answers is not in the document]
VasylMarchuk: > just vercel's caching config should work

That's what I'm unsure about, need to check, vercel caching config might only influence how soon the function is re-triggered when requested, and if it isn't, it just serves a response cached in vercel. We might need to add actual caching headers for the users' browsers to also cache the response.

vercel-functions/prerender-user-profile-page.prerender-config.json
```diff
@@ -0,0 +1,5 @@
+{
+  "expiration": 60,
+  "allowQuery": ["username"],
```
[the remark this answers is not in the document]
VasylMarchuk: So yeah, it's part of the _path_ **in the route**, but the serverless function receives dynamic path segments as `queryParams`, while it's own `request.url` is the URL of the serverless function itself:

So if I have this in `vercel.json`:

```json
{ "source": "/users/:userName/:remainingPath*", "destination": "/synchrophasotron9000" },
```

Then the function receives this info in the `request`:

Знімок екрана 2025-02-07 о 16 58 46

vercel.json
```diff
@@ -1,33 +1,21 @@
...
+    { "source": "/", "destination": "/catalog" },
+    { "source": "/join_team", "destination": "https://backend.codecrafters.io/join_team" },
+    { "source": "/setup_team", "destination": "https://backend.codecrafters.io/setup_team" },
+    { "source": "/r/famous-wombat-417894", "destination": "/join?via=gnomezgrave" }
+  ],
   "rewrites": [
     { "source": "/submissions", "destination": "https://backend.codecrafters.io/submissions" },
     { "source": "/submissions/:match*", "destination": "https://backend.codecrafters.io/submissions/:match*" },
     { "source": "/progress/:match*", "destination": "https://backend.codecrafters.io/progress/:match*" },
+    { "source": "/users/:username", "destination": "/prerender-user-profile-page" },
```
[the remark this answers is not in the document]
VasylMarchuk: Alright, will do, this is also a package name used in the function's `package.json`, and so it has to be a valid NPM package name too.

package.json
VasylMarchuk: Great suggestion, I will just invoke the scripts directly in `vercel.json`! These `build:vercel:dist` and `build:vercel:functions` tasks are a remnant of the previous idea, I'll remove them!

But can I at least keep separate `build:versions` and `build:ember` tasks, both called by `build` @rohitpaulk ?

This is useful as I improved `build:versions` a little bit to show more info, and ALSO, I have an experimental branch where I enable [fastboot rehydration](https://github.com/ember-fastboot/ember-cli-fastboot?tab=readme-ov-file#rehydration) and need to add `EXPERIMENTAL_RENDER_MODE_SERIALIZE=true` to `ember build` — and this all turns the `build` task into a mess.

So I just wanted to have a separate task for `ember build` and nothing else.

scripts/build-vercel-dist.sh
VasylMarchuk: Great idea

scripts/build-vercel-functions.sh
VasylMarchuk: Awesome tip, I completely forgot about those, didn't do much bash lately!

vercel-functions/prerender-user-profile-page.func/index.js
VasylMarchuk: So unfortunately no, I tried this explicitly while experimenting, and the middleware does NOT run after a serverless function — it's either the function or the middleware who handle the route (the function takes precedence).

BUT! In this PR, I removed handling of users route from the middleware, and in my next one, I will remove handling concepts, and then — the only remaining thing in the middleware would be contests.

And I thought that we'd just move contests pre-rendering to the serverless function too, and then remove the middleware completely. Makes sense @rohitpaulk ?

vercel-functions/prerender-user-profile-page.func/index.js
VasylMarchuk: I loved this idea [at first], been thinking about it for a couple hours.

Wondering if it's possible to redirect multiple routes to a function, for example:

```
/users/:suffix --> /prerender-ember-route?prefix=/users/
/concepts/:suffix --> /prerender-ember-route?prefix=/concepts/
```

and have the function receive `prefix` and `suffix` as query parameters.

I'm not sure if vercel's routing can do that (the custom `?prefix=...` part in the `destination`), will check if it can.

vercel.json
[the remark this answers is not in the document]
VasylMarchuk: But if we use a single function for multiple routes as discussed below — this will be moot, as the function would just be called something like `/prerender-ember-route`

vercel-functions/prerender-user-profile-page.func/index.js
VasylMarchuk: This ^ is the best idea so far, I think, directory structure would look like:

```
functions/
  prerender/
    _prerender_ember_route.func/
    users/
      user.func <-- symlink to `../_prerender_ember_route.func`
    concepts/
      concept.func <-- symlink to `../_prerender_ember_route.func`
```

Direct requests to `/prerender/**` would be redirected to `/404`

Requests to `/users/:id` would be rewritten to `/prerender/users/user?id=...`
Requests to `/concepts/:id` would be rewritten to `/prerender/concept/contest?id=...`

And the function `_prerender_ember_route` would be able to tell the ember path to prerender from the alias:

```
/prerender/users/user/:id --> visit `/users/user/:id`
/prerender/concepts/concept/:id --> visit `/concepts/concept/:id`
```

app/routes/user.ts
```diff
@@ -13,23 +15,65 @@ export default class UserRoute extends BaseRoute {
...
+  }
+
+  async #fetchUserRecord(username: string) {
+    return (
+      (await this.store.query('user', {
+        username,
+        include: 'course-participations.language,course-participations.course.stages,course-participations.current-stage,profile-events',
+      })) as unknown as UserModel[]
+    )[0];
+  }
```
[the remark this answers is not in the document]
VasylMarchuk: `records[0]` will be `undefined` already if `records` is an empty array, no need to do it explicitly

app/routes/user.ts
```diff
@@ -13,23 +14,65 @@ export default class UserRoute extends BaseRoute {
...
```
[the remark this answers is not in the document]
VasylMarchuk: I already responded to this: `records[0]` will be `undefined` already if `records` is an empty array, no need to do it explicitly

vercel-functions/prerender-user-profile-page.func/index.js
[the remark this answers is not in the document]
VasylMarchuk: So it seems that we're currently not setting any **browser-side** caching headers at all, neither for our static assets not for pre-rendered pages like `/catalog`, so even if we do want to set them — it's a matter of a separate PR:

Знімок екрана 2025-02-14 о 15 50 17
Знімок екрана 2025-02-14 о 15 50 46

Closing this conversation for now

vercel.json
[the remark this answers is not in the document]
VasylMarchuk: Since the function has been renamed to `/prerender/_prerender_ember_route` and the actual route functions are aliases — this is now moot.

vercel-functions/prerender-user-profile-page.func/index.js
VasylMarchuk: Converted to the symlinks approach

vercel.json
```diff
@@ -1,33 +1,21 @@
...
+    { "source": "/", "destination": "/catalog" },
+    { "source": "/join_team", "destination": "https://backend.codecrafters.io/join_team" },
+    { "source": "/setup_team", "destination": "https://backend.codecrafters.io/setup_team" },
+    { "source": "/r/famous-wombat-417894", "destination": "/join?via=gnomezgrave" }
+  ],
   "rewrites": [
     { "source": "/submissions", "destination": "https://backend.codecrafters.io/submissions" },
     { "source": "/submissions/:match*", "destination": "https://backend.codecrafters.io/submissions/:match*" },
     { "source": "/progress/:match*", "destination": "https://backend.codecrafters.io/progress/:match*" },
+    { "source": "/users/:suffix", "destination": "/prerender/users/user" },
```
rohitpaulk, line 16: The destination path convention is not super clear to me here, can we have it be based on the ember route name directly? i.e. `/prerender/user`.

vercel-functions/prerender/_prerender-ember-route.func/index.js
```diff
@@ -0,0 +1,63 @@
...
+    query: { suffix },
+  } = request;
+
+  let prefix = (url || '/')
+    .replace(/\?(.*)$/, '') // strip query parameters
+    .replace(/^\/prerender\//, '/') // strip /prerender
+    .replace(/([^^])\/$/, '$1'); // strip trailing /
+
+  if (suffix) {
+    prefix = prefix.replace(/((\/[^/]+))$/, '') || '/'; // strip wildcard function name
```
rohitpaulk, line 16: @VasylMarchuk I'm kinda confused with this - can you explain a bit more what's going on here? What's a "wildcard function name"?

rohitpaulk: One tip if it helps: I've found it useful to include a brief description of the inputs / outputs of a function that does parsing/transformation. Something like: 

```
// 'https://google.com/path?query=value' -> '/path'
function extractPath() {}
```

vercel.json
rohitpaulk: I wonder if a way to be more explicit here would be to do:

```suggestion
    { "source": "/users/:suffix", "destination": "/prerender/user?prefix=/users" },
```

That way it's more clear what "prefix" and "suffix" mean, and the naming convention is more clearly based on ember route names

vercel-functions/prerender/_prerender-ember-route.func/index.js
VasylMarchuk: Will add a JSDoc comment!

vercel.json
VasylMarchuk: So the trick is, I want to function to be able to handle both types of routes:

```
/users/:suffix --> resolves to /users/user.func — wildcard function that gets passed `suffix` as a queryParam
/demo/code-mirror --> resolves to /demo/code-mirror.func — no wildcard, no suffix, just render a direct route
```

I can add better names for suffix & prefix, for example `_prerenderModelId` and `prerenderPrefix`, and add JSDoc to the `extractPath` function.

This approach is better than passing `prefix=...` manually, because it works automatically and you don't have to pass the prefix.

rohitpaulk: Oh, okay - yeah got it. So the convention is just based on paths then (instead of ember routes), makes sense.

Let's go ahead and merge this, will think about an alternative that is slightly less confusing - still think the prefix/suffix stuff is pretty indirect and could be cleaned up. Nothing comes to mind right now but I'm sure there's a way out here...

rohitpaulk (approved):

vercel.json
VasylMarchuk: Btw, I chose `prefix` and `suffix` because in theory there should also be `ending`: 

If I want to rewrite `/demo/code-mirror?param1&param2&param3&...`, to my prerender function and pass those query params to `visit()` — then `ending` wound be the query param part: `?param1&...`

vercel-functions/prerender/_prerender-ember-route.func/index.js
VasylMarchuk: Moved the function into a separate module and added JSDoc

vercel.json
VasylMarchuk: Found a better approach, using function names wrapped in `[]`, for example `/prerender/users/[user].func`, to denote "wildcard" functions, as done [in this vercel example](https://github.com/vercel/examples/tree/main/build-output-api/on-demand-isr/.vercel/output/functions/blog). Updated `parsePrerenderPath` to reflect this.

vercel-functions/prerender/_prerender-ember-route.func/parse-prerender-path.js
```diff
@@ -0,0 +1,76 @@
...
+  /test -> /prerender/test
+  ```
+  Making requests below will result in corresponding Request URLs received by
+  the Prerender Function, and `parsePrerenderPath` will produce the following
+  FastBoot paths for such Request URLs:
+  ```
+  /demo/code-mirror -> /prerender/demo/code-mirror -> /demo/code-mirror
+  /users/abcde -> /prerender/users/[user]?user=abcde -> /users/abcde
+  /concepts/xyzqwe -> /prerender/concepts/[concept]?concept=xyzqwe -> /concepts/xyzqwe
+  /examples/abc/def -> /prerender/examples/[page][subpage]?page=abc&subpage=def -> /examples/abc/def
```
rohitpaulk, line 32:
```suggestion
  /examples/abc/def -> /prerender/examples/[page]/[subpage]?page=abc&subpage=def -> /examples/abc/def
```

(Let me know if this is incorrect)

rohitpaulk: Oh, okay I see what's happening here - the last part **needs** to not have the `/` for this to work, `parsePrerenderPath` assumes this.

rohitpaulk (approved):

rohitpaulk:
@VasylMarchuk can you check what's up with this user? 

https://app.codecrafters.io/users/Gufran (this works) 
https://frontend-4k45tcyj7.ccio.dev/users/Gufran (this returns a 404)

rohitpaulk:
I wonder if it's the capitalization causing issues?

vercel-functions/prerender/_prerender-ember-route.func/parse-prerender-path.js
VasylMarchuk: Exactly! As far as I understood a vercel `something.func` can not contain any more nested `other.func` functions

middleware.js
```diff
@@ -95,21 +87,17 @@ export default async function middleware(request) {
...
+      const conceptData = await (
+        await fetch(`https://backend.codecrafters.io/services/dynamic_og_images/concept_data?id_or_slug=${conceptSlug}`)
+      ).json();
       conceptTitle = conceptData.title;
       conceptDescription = conceptData.description_markdown;
     } catch (e) {
-      console.error(e);
-      console.log('ignoring error for now');
+      console.error('Failed to fetch concept data:', e);
     }
```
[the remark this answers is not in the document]
VasylMarchuk: This PR is not touching neither Contests nor Concepts, only Users, and the changes in the middleware are purely formatting-related, and will be addressed in subsequent PRs

PR #2660, opened 2025-02-18
Fix focus highlighting for question cards

ryan-gang:
Improve keyboard navigation by adding support for keyboard events, managing navigation state in parent components, and ensuring proper focus styling. This update enhances user experience when navigating options using the keyboard.

**Checklist**:


ryan-gang:
We opted for a no state-management-required solution.
Merged: https://github.com/codecrafters-io/frontend/pull/2668

PR #2674, opened 2025-02-21
Refactor `authenticator.initiateLogin`

ryan-gang:
Update multiple components to call initiateLogin without a null parameter, simplifying login initiation across the application

**Checklist**:

## Summary by CodeRabbit

- **New Features**
	- Enhanced the login experience by automatically redirecting users to their intended destinations after signing in.
	- Introduced a new method for handling login initiation with redirection capabilities.

- **Refactor**
	- Streamlined authentication calls across various pages by removing unnecessary parameters, simplifying the login process.

These updates result in a smoother and more cohesive login process on several user-facing pages, improving overall navigation and reducing friction during authentication.

### DIFF
--- a/app/services/authenticator.ts
+++ b/app/services/authenticator.ts
@@ -81,12 +81,14 @@ export default class AuthenticatorService extends Service {
   }
 
   initiateLoginAndRedirectTo(redirectPathOrUrl: string) {
-    if (redirectPathOrUrl.startsWith(`${config.x.backendUrl}`)) {
-      window.location.href = `${config.x.backendUrl}/login?next=` + redirectPathOrUrl;
-    } else if (redirectPathOrUrl.startsWith('/')) {
-      window.location.href = `${config.x.backendUrl}/login?next=` + encodeURIComponent(`${window.origin}${redirectPathOrUrl}`);
+    const loginBaseUrl = `${config.x.backendUrl}/login?next=`;
+
+    if (redirectPathOrUrl.startsWith(config.x.backendUrl)) {
+      window.location.href = loginBaseUrl + redirectPathOrUrl;
     } else {
-      window.location.href = `${config.x.backendUrl}/login?next=` + encodeURIComponent(redirectPathOrUrl);
+      const fullRedirectUrl = redirectPathOrUrl.startsWith('/') ? `${window.origin}${redirectPathOrUrl}` : redirectPathOrUrl;
+
+      window.location.href = loginBaseUrl + encodeURIComponent(fullRedirectUrl);
     }
   }
 
--- a/app/services/authenticator.ts
+++ b/app/services/authenticator.ts
@@ -72,13 +72,7 @@ export default class AuthenticatorService extends Service {
     await this.syncCurrentUser();
   }
 
-  initiateLogin(redirectPathOrUrl: string | null = null) {
-    if (redirectPathOrUrl) {
-      this.initiateLoginAndRedirect(redirectPathOrUrl);
-
-      return;
-    }
-
+  initiateLogin() {
     if (this.router.currentURL) {
       window.location.href = `${config.x.backendUrl}/login?next=` + encodeURIComponent(`${window.origin}${this.router.currentURL}`);
     } else {
@@ -86,7 +80,7 @@ export default class AuthenticatorService extends Service {
     }
   }
 
-  initiateLoginAndRedirect(redirectPathOrUrl: string) {
+  initiateLoginAndRedirectTo(redirectPathOrUrl: string) {
     if (redirectPathOrUrl.startsWith(`${config.x.backendUrl}`)) {
       window.location.href = `${config.x.backendUrl}/login?next=` + redirectPathOrUrl;
     } else if (redirectPathOrUrl.startsWith('/')) {
--- a/app/services/authenticator.ts
+++ b/app/services/authenticator.ts
@@ -74,16 +74,16 @@ export default class AuthenticatorService extends Service {
 
   initiateLogin() {
     if (this.router.currentURL) {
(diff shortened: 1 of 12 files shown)


app/services/authenticator.ts
```diff
@@ -72,20 +72,30 @@ export default class AuthenticatorService extends Service {
...
 
-  initiateLogin(redirectPath: string | null) {
-    const isFullUrl = redirectPath?.startsWith(`${config.x.backendUrl}`);
-
-    if (isFullUrl) {
-      window.location.href = `${config.x.backendUrl}/login?next=` + redirectPath;
-    } else if (redirectPath) {
-      window.location.href = `${config.x.backendUrl}/login?next=` + encodeURIComponent(`${window.origin}${redirectPath}`);
-    } else if (this.router.currentURL) {
+  initiateLogin(redirectPathOrUrl: string | null = null) {
```
rohitpaulk, line 75:
```suggestion
  initiateLogin() {
```

Let's remove the optional parameter here - let callers call `initiateLoginAndRedirectTo()` if needed. 

The intention here is to simplify the interface, and not have too many branches that the caller needs to reason about - by adding this branch we haven't really made this any simpler.

app/services/authenticator.ts
```diff
@@ -72,20 +72,30 @@ export default class AuthenticatorService extends Service {
...
+    }
+
+    if (this.router.currentURL) {
       window.location.href = `${config.x.backendUrl}/login?next=` + encodeURIComponent(`${window.origin}${this.router.currentURL}`);
     } else {
       window.location.href = `${config.x.backendUrl}/login?next=` + encodeURIComponent(window.origin);
     }
   }
 
+  initiateLoginAndRedirect(redirectPathOrUrl: string) {
```
rohitpaulk, line 89:
```suggestion
  initiateLoginAndRedirectTo(redirectPathOrUrl: string) {
```o

Lightly suggestive of what the argument is, think this'll read better to

File: app/services/authenticator.ts (hunk 1 above)
Comment (ryan-gang) [reply app/services/authenticator.ts:75]: I was thinking exposing a single function was better, so we could internally handle all the complexity. And not make callers decide which one to call.

File: app/services/authenticator.ts [hunk 3]
```diff
@@ -72,20 +72,30 @@ export default class AuthenticatorService extends Service {
...
+    if (redirectPathOrUrl.startsWith(`${config.x.backendUrl}`)) {
+      window.location.href = `${config.x.backendUrl}/login?next=` + redirectPathOrUrl;
+    } else if (redirectPathOrUrl.startsWith('/')) {
+      window.location.href = `${config.x.backendUrl}/login?next=` + encodeURIComponent(`${window.origin}${redirectPathOrUrl}`);
```
rohitpaulk, line 93: Still too much branching - @ryan-gang can you take a fresh look at this and try to simplify? I'm positive this whole thing can be made far simpler than it currently is. 

Primary thing that stands out here is: Why should backend URLs be treated any differently? We only need to differentiate between "path" and "URL"

rohitpaulk (changes requested):
Doesn't really simplify things - let's take another pass

app/services/authenticator.ts
rohitpaulk: Hmm, then why have two functions at all? Just keep one (and make the other private?)?

That's fine by me too, would be a simple enough interface

rohitpaulk: @ryan-gang one way to think about this is "intention" - 90% of our cases, the intention is just to initiate login with the current URL as the redirect target. It's only in the other 10% where we need to redirect a user to a specific location. Given that split, keeping two functions could make sense.

app/services/authenticator.ts
```diff
@@ -72,20 +72,20 @@ export default class AuthenticatorService extends Service {
...
-    const isFullUrl = redirectPath?.startsWith(`${config.x.backendUrl}`);
-
-    if (isFullUrl) {
-      window.location.href = `${config.x.backendUrl}/login?next=` + redirectPath;
-    } else if (redirectPath) {
-      window.location.href = `${config.x.backendUrl}/login?next=` + encodeURIComponent(`${window.origin}${redirectPath}`);
-    } else if (this.router.currentURL) {
+  initiateLogin() {
+    if (this.router.currentURL) {
       window.location.href = `${config.x.backendUrl}/login?next=` + encodeURIComponent(`${window.origin}${this.router.currentURL}`);
```
rohitpaulk, line 77:
```suggestion
      this.initiateLoginAndRedirectTo(`${window.origin}${this.router.currentURL}`);
```

app/services/authenticator.ts
```diff
@@ -72,20 +72,20 @@ export default class AuthenticatorService extends Service {
...
     } else {
       window.location.href = `${config.x.backendUrl}/login?next=` + encodeURIComponent(window.origin);
```
rohitpaulk, line 79:
```suggestion
      this.initiateLoginAndRedirectTo(window.origin);
```

app/services/authenticator.ts
```diff
@@ -72,20 +72,20 @@ export default class AuthenticatorService extends Service {
...
     }
   }
 
+  initiateLoginAndRedirectTo(redirectPathOrUrl: string) {
+    const fullRedirectUrl = redirectPathOrUrl.startsWith('/') ? `${window.origin}${redirectPathOrUrl}` : redirectPathOrUrl;
```
rohitpaulk, line 84:
```suggestion
    const redirectUrl = redirectPathOrUrl.startsWith('/') ? `${window.origin}${redirectPathOrUrl}` : redirectPathOrUrl;
```

("url" vs "path" is already clear, so just `redirectUrl` should do here? Another option is `finalRedirectUrl` to signify that there's an intermediary redirect followed by another. "Full" doesn't add much meaning)

rohitpaulk (approved):

PR #2670, opened 2025-02-20
Prerender Concept pages with Vercel Prerender Functions

VasylMarchuk:
Second part of #2607

### Brief

This adds server-side rendering for `/concepts/:concept` route, based on Vercel's [Prerender Functions](https://vercel.com/docs/build-output-api/v3/primitives#prerender-functions).

### Details

- Set proper OG meta tags for `/concepts/concept` route
- Added an alias function `/prerender/concepts/[concept]` which reuses `/prerender/_prerender-ember-route` function
- Made `/prerender/concepts/[concept]` function handle `/concepts/:concept` route instead of the middleware
  - `expiration` time is currently set to `24 hours`
- Removed `/concepts/:concept` route handling from the middleware and cleaned it up
  - The middleware is now handling only _contests_
- Fixed a couple FastBoot related undefined reference errors

### Checklist

- [x] I've thoroughly self-reviewed my changes
- [x] I've added tests for my changes, unless they affect admin-only areas (or are otherwise not worth testing)
- [x] I've verified any visual changes using Percy (add a commit with `[percy]` in the message to trigger)

## Summary by CodeRabbit

- **New Features**
  - Concept pages now feature dynamic metadata updates and enhanced prerendering for improved performance and SEO.
  - Contest-related routes have been streamlined for more efficient processing.
  - New configuration for prerendering allows query parameters to be included.

- **Bug Fixes**
  - URL parameter handling has been improved to prevent runtime errors in non-browser environments.

PR #2694, opened 2025-02-22
Fix redirect to 404 from concept route when concept is not found

VasylMarchuk:
Related to #2607
Follow-up to https://github.com/codecrafters-io/frontend/pull/2670

### Brief

This fixes the redirection to `/404` from the `/concepts/concept` route when the requested Concept could not be found

### Details

- Moved the redirection into the `afterModel` hook
  - it redirects to `404` if the `model` hook returned `undefined`
  - the `model` hook returns `undefined` if Concept doesn't exist
  - doing the redirect directly in the `model` hook causes a `TransactionAborted` error
- Added a test for redirection of non-existing concepts to `/404`
- Renamed `findOrCreateConceptEngagement` to `#findCachedOrCreateNewConceptEngagement`

### Screenshot

Знімок екрана 2025-02-22 о 12 14 34

### Checklist

## Summary by CodeRabbit

- **New Features**
  - Enhanced navigation by adding a method to visit specific concept pages using dynamic URLs.
  
- **Bug Fixes**
  - Improved error handling for missing content to provide a smoother navigation experience.
  - Adjusted data loading so that engagement information is only fetched for authenticated users.
  - Added a test to ensure proper redirection to a 404 error page for non-existing concepts.

### DIFF
--- a/app/routes/concept.ts
+++ b/app/routes/concept.ts
@@ -37,7 +37,15 @@ export default class ConceptRoute extends BaseRoute {
     });
   }
 
-  afterModel({ concept }: ConceptRouteModel): void {
+  afterModel(model: ConceptRouteModel): void {
+    if (!model) {
+      this.router.transitionTo('not-found');
+
+      return;
+    }
+
+    const { concept } = model;
+
     // Save previous OG meta tags
     this.previousMetaImageUrl = this.metaData.imageUrl;
     this.previousMetaTitle = this.metaData.title;
@@ -64,7 +72,7 @@ export default class ConceptRoute extends BaseRoute {
     this.metaData.description = this.previousMetaDescription;
   }
 
-  async findOrCreateConceptEngagement(concept: ConceptModel) {
+  async #findCachedOrCreateNewConceptEngagement(concept: ConceptModel) {
     const latestConceptEngagement = this.authenticator.currentUser?.conceptEngagements
       .filter((engagement) => engagement.concept.slug === concept.slug)
       .sortBy('createdAt')
@@ -86,9 +94,7 @@ export default class ConceptRoute extends BaseRoute {
     const concept = allConcepts.find((concept) => concept.slug === params.concept_slug);
 
     if (!concept) {
-      this.router.transitionTo('not-found');
-
-      return;
+      return; // will redirect to 404 in afterModel
     }
 
     const allConceptGroups = await this.store.findAll('concept-group');
@@ -102,7 +108,7 @@ export default class ConceptRoute extends BaseRoute {
       });
     }
 
-    const latestConceptEngagement = await this.findOrCreateConceptEngagement(concept);
+    const latestConceptEngagement = await this.#findCachedOrCreateNewConceptEngagement(concept);
 
     return {
       allConcepts,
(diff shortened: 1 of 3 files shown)


app/routes/concept.ts
```diff
@@ -65,6 +73,12 @@ export default class ConceptRoute extends BaseRoute {
   }
 
   async findOrCreateConceptEngagement(concept: ConceptModel) {
+    if (this.authenticator.isAuthenticated) {
```
rohitpaulk, line 76: @VasylMarchuk is this `findOrCreate` used in multiple places? Want to know if there's a specific reason to move the `findAll` into here. 

We generally try to keep all external API calls at the highest level if possible (the model hook in this case), just makes it more obvious what all endpoints a route hits.

rohitpaulk (changes requested):
Q added

app/routes/concept.ts
VasylMarchuk: Well, I didn't write this function, someone else probably did just to extract this piece into a logical and well-named function. And he probably forgot to move this `findAll` from the model hook along with the rest of the logic.

That findAll belongs in `findOrCreateConceptEngagement`, it makes no sense in the model hook itself, and `findOrCreateConceptEngagement` doesn't make complete sense without it.

So I just moved it:
Знімок екрана 2025-02-26 о 10 42 40

That said, this method is not used anywhere else, and I wouldn't mind just getting rid of it and moving the logic back into the model hook, would you like me to do that @rohitpaulk ?

VasylMarchuk: If anything, I'd rename that method to `#findOrCreateConceptEngagement`, to make it private.

rohitpaulk: @VasylMarchuk - yep, let's move back and rename to private!

rohitpaulk: > That findAll belongs in findOrCreateConceptEngagement, it makes no sense in the model hook itself

Just for context, that method is supposed to be "finding" records that are already loaded locally - not via a network request. I see how the naming is confusing, but that's what the intention is. 

Route's responsibility - load remote records
Function's responsibility - load record from store or create record in store

Must be a better name for this, no great ideas at the moment though. Maybe `findOrCreateLocalConceptEngagement`?

VasylMarchuk: findLoadedOrCreateNewConceptEngagement?

VasylMarchuk: Done, renamed to `#findCachedOrCreateNewConceptEngagement`

app/routes/concept.ts
```diff
@@ -37,7 +37,15 @@
     });
   }
 
-  afterModel({ concept }: ConceptRouteModel): void {
+  afterModel(model: ConceptRouteModel): void {
+    if (!model) {
+      this.router.transitionTo('not-found');
+
+      return;
```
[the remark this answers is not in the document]
VasylMarchuk: Test added

rohitpaulk (approved):

tests/pages/concepts-page.ts
```diff
@@ -29,4 +29,5 @@ export default createPage({
   helpscoutBeacon: helpscoutBeacon,
 
   visit: visitable('/concepts'),
+  visitConcept: visitable('/concepts/:concept_slug'),
```
rohitpaulk, line 32: Note: this'd usually go into `concept-page.ts` (not `concepts-page.ts`)

PR #2722, opened 2025-03-04
Fix blinking on Concept route for logged in users

### LINKED ISSUE #2607: Statically rendered user pages + concept pages

VasylMarchuk:
Closes #2607 

### Brief

This fixes the blinking on the Concept page caused by a blocking `findAll` in the route's model hook, when trying to fetch the Concept Engagement for logged in users.

### Details

- Add a `700ms` delay before scrolling down to the congratulations icon after completing a Concept
   - Added a `delay` argument to `app/modifiers/scroll-into-view`, by default `0`
- Better handling of `@latestConceptEngagement` argument in `app/components/concept`
- Move `latestConceptEngagement` from `ConceptRouteModel` into `ConceptController`, so that it can be `@tracked`
  - move setting up `latestConceptEngagement` from the route's `model` hook into `setupController` hook
- Better logic for finding the concept in the Concept route model hook
  - it tries to find the concept in the store cache (where it could be inserted from Shoebox)
  - and if it can't find the concept — it tries fetching it from the backend with `reload: true`
  - and only then it redirects to `404` if concept is not found
  - this fixes both the blinking, and visiting `/concepts/:slug` route from other `link-to` helpers
- Better logic for finding concept engagement model
  - it will return an existing engagement if one is cached in the store
  - it will create and return a new engagement otherwise
  - it will also schedule re-fetching of the engagement in the background and update the controller's `latestConceptEngagement` accordingly afterwards
- Fix the `visit` helper in `ConceptPage` page object
- Remove `visitConcept` helper from `ConceptsPage` page object

### Checklist

## Summary by CodeRabbit

- **New Features**
  - Introduced smoother scrolling on concept pages with a brief delay, resulting in a more fluid transition when new content is revealed.
  - Updated the URL structure for accessing concept pages, providing a more intuitive and consistent navigation experience.

- **Refactor**
  - Enhanced the management of user engagement data to ensure that progress tracking remains accurate and responsive during interactions.
  - Streamlined the logic for determining the current block group index, improving performance and clarity.
  - Improved the structure of acceptance tests for visiting concept pages, consolidating functionality for better maintainability.
  - Updated property bindings in templates for better data retrieval and usage.

### DIFF
--- a/app/components/concept/index.hbs
+++ b/app/components/concept/index.hbs
@@ -1,9 +1,4 @@
-<div
-  {{did-insert this.handleDidInsertContainer}}
-  {{did-update this.recalculateLastRevealedBlockGroupIndex @latestConceptEngagement}}
-  {{will-destroy this.handleWillDestroyContainer}}
-  ...attributes
->
+<div {{did-insert this.handleDidInsertContainer}} {{will-destroy this.handleWillDestroyContainer}} ...attributes>
   {{#each this.visibleBlockGroups as |blockGroup|}}
     <div
       class="scroll-mb-10 md:scroll-mb-[20vh] scroll-mt-20 md:scroll-mt-[20vh]
--- a/app/components/concept/index.hbs
+++ b/app/components/concept/index.hbs
@@ -1,4 +1,9 @@
-<div {{did-insert this.handleDidInsertContainer}} {{will-destroy this.handleWillDestroyContainer}} ...attributes>
+<div
+  {{did-insert this.handleDidInsertContainer}}
+  {{did-update this.recalculateLastRevealedBlockGroupIndex @latestConceptEngagement}}
+  {{will-destroy this.handleWillDestroyContainer}}
+  ...attributes
+>
   {{#each this.visibleBlockGroups as |blockGroup|}}
     <div
       class="scroll-mb-10 md:scroll-mb-[20vh] scroll-mt-20 md:scroll-mt-[20vh]
--- a/app/routes/concept.ts
+++ b/app/routes/concept.ts
@@ -4,37 +4,19 @@ import { inject as service } from '@ember/service';
 import { scheduleOnce } from '@ember/runloop';
 import ConceptModel from 'codecrafters-frontend/models/concept';
 import ConceptGroupModel from 'codecrafters-frontend/models/concept-group';
-import ConceptEngagementModel from 'codecrafters-frontend/models/concept-engagement';
 import AuthenticatorService from 'codecrafters-frontend/services/authenticator';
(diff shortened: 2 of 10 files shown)


app/routes/concept.ts
```diff
@@ -10,13 +10,31 @@ import type MetaDataService from 'codecrafters-frontend/services/meta-data';
...
+import { tracked } from '@glimmer/tracking';
 
-export type ConceptRouteModel = {
+export class ConceptRouteModel {
   allConcepts: ConceptModel[];
   concept: ConceptModel;
   conceptGroup: ConceptGroupModel;
-  latestConceptEngagement: ConceptEngagementModel;
-};
+  @tracked latestConceptEngagement: ConceptEngagementModel;
```
rohitpaulk, line 19: This is pretty non-standard - can we move the tracked property to be in the controller instead somehow?

rohitpaulk (changes requested):
Added note re: how using a class for the route model is very non-standard

app/components/concept/index.hbs
```diff
@@ -1,4 +1,9 @@
-<div {{did-insert this.handleDidInsertContainer}} {{will-destroy this.handleWillDestroyContainer}} ...attributes>
+<div
+  {{did-insert this.handleDidInsertContainer}}
+  {{did-update this.recalculateLastRevealedBlockGroupIndex @latestConceptEngagement}}
```
rohitpaulk, line 3: @VasylMarchuk shouldn't this just work by default? We're using `latestConceptEngagement.currentProgressPercentage` inside a function, as long as that is tracked, the value should update automatically no?

app/routes/concept.ts
VasylMarchuk: Yes, I tried that initially, it needs to be a controller property, and it needs to be set up using the `setupController` hook, I can do that, it just seemed a bit messier, more code than just changing `type` -> `class`

VasylMarchuk: I have reverted `ConceptRouteModel` back to being a `type`, and moved `latestConceptEngagement` property from the model into the `ConceptController`, where it is naturally `@tracked`.

I've also moved the logic of assigning `latestConceptEngagement` from the `model` hook into the `setupController` hook in the route.

rohitpaulk (approved):

PR #2745, opened 2025-03-24
Update frontend to use `Course#visibility` attribute to selectively render courses

ryan-gang:
Introduce a visibility attribute to the CourseModel and ensure that private courses are filtered out from participation groups, course listings, and extension ideas. This enhances access control and user experience by preventing private courses from appearing in relevant contexts.

**Checklist**:

## Summary by CodeRabbit

- **New Features**
  - Introduced a visibility setting for courses, allowing them to be marked as public or private.
  - Enhanced the filtering logic across multiple areas to exclude courses marked as private—along with those already restricted due to other statuses—improving the relevance of displayed course lists.

- **Bug Fixes**
  - Ensured private courses are not displayed in various contexts, including user profiles and course catalogs.

- **Tests**
  - Implemented new acceptance tests to verify the correct handling of private courses in the track page, course catalog, and user profile views.

### DIFF
--- a/app/models/course.ts
+++ b/app/models/course.ts
@@ -44,6 +44,7 @@ export default class CourseModel extends Model {
   @attr('string') declare shortName: string;
   @attr('string') declare slug: string;
   @attr('string') declare testerRepositoryFullName: string;
+  @attr('string') declare visibility: 'public' | 'private';
 
   @attr() declare testimonials: {
     author_name: string;
--- a/app/models/course.ts
+++ b/app/models/course.ts
@@ -81,6 +81,9 @@ export default class CourseModel extends Model {
   @equal('releaseStatus', 'live') declare releaseStatusIsLive: boolean;
   @equal('releaseStatus', 'deprecated') declare releaseStatusIsDeprecated: boolean;
 
+  @equal('visibility', 'public') declare visibilityIsPublic: boolean;
+  @equal('visibility', 'private') declare visibilityIsPrivate: boolean;
+
   @service declare date: DateService;
 
   get baseStages() {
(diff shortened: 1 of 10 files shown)


app/models/course.ts
```diff
@@ -44,6 +44,7 @@ export default class CourseModel extends Model {
   @attr('string') declare shortName: string;
   @attr('string') declare slug: string;
   @attr('string') declare testerRepositoryFullName: string;
+  @attr('string') declare visibility: 'public' | 'private';
```
rohitpaulk, line 47: Let's add accessors like `visibilityIsPrivate` and `visibilityIsPublic` and use them in the other places where we're accessing this. Keeps things consistent with `releaseStatusIs...` and prevents any mistakes in using the enum value. 

Code like `rejectBy('visibility', 'prvate')` won't cause a type error but `rejectBy('visibilityIsPrvate')` will

tests/acceptance/track-page/view-track-test.js
```diff
@@ -140,6 +140,34 @@ module('Acceptance | track-page | view-track', function (hooks) {
...
+    redis.update({ visibility: 'private' });
+
+    this.server.create('repository', {
+      course: redis,
+      language: python,
+      user: currentUser,
+    });
+
+    await visit('/tracks/python');
+    assert.notOk(trackPage.cards.mapBy('title').includes('Build your own Redis'), 'private course should not be included');
```
rohitpaulk, line 168: Umm in this case it _should_ be visible no? Since the user has a repository?

rohitpaulk: Oh ignore, this is the track page

rohitpaulk (approved):

PR #2780, opened 2025-04-06
Fix course comments with code blocks extending past container

### LINKED ISSUE #1591: Styling issue: Comments with code blocks extend past container
From https://app.codecrafters.io/courses/http-server/stages/1: 

Screenshot 2024-04-12 at 14 25 05

(These comments might not be visible, so you'll need to create your own to actually replicate the issue)

VasylMarchuk:
Closes #1591

### Brief

This fixes Course comments with code blocks extending past their container's width.

### Details

After adding `min-w-0` to the comment cards, the sizing logic starts working properly.

### Demo

https://github.com/user-attachments/assets/7b92ac4d-a0a7-477b-a250-d3951d228d4a

### Checklist

## Summary by CodeRabbit

- **Style**
  - Improved layout consistency for comment components by adding a new CSS class to enhance their appearance.

### DIFF
--- a/app/components/course-page/comment-timeline-item.hbs
+++ b/app/components/course-page/comment-timeline-item.hbs
@@ -1,6 +1,6 @@
-<div class="relative pl-11">
-  <a href={{@author.codecraftersProfileUrl}} class="absolute left-0 top-0.5 flex items-center" target="_blank" rel="noopener noreferrer">
-    <AvatarImage @user={{@author}} class="w-8 h-8 filter drop-shadow-sm ring-1 ring-white dark:ring-white/5 rounded-full shadow" />
+<div class="flex items-start">
+  <a href={{@author.codecraftersProfileUrl}} class="mt-0.5 flex items-center flex-shrink-0" target="_blank" rel="noopener noreferrer">
+    <AvatarImage @user={{@author}} class="w-8 h-8 mr-3 filter drop-shadow-sm ring-1 ring-white dark:ring-white/5 rounded-full shadow" />
   </a>
 
   {{yield}}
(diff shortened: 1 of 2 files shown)


app/components/course-page/comment-timeline-item.hbs
```diff
@@ -1,6 +1,6 @@
-<div class="flex items-start">
-  <a href={{@author.codecraftersProfileUrl}} class="mt-0.5 flex items-center flex-shrink-0" target="_blank" rel="noopener noreferrer">
-    <AvatarImage @user={{@author}} class="w-8 h-8 mr-3 filter drop-shadow-sm ring-1 ring-white dark:ring-white/5 rounded-full shadow" />
+<div class="relative pl-11">
+  <a href={{@author.codecraftersProfileUrl}} class="absolute left-0 top-0.5 flex items-center" target="_blank" rel="noopener noreferrer">
```
rohitpaulk, line 2: @VasylMarchuk is there no cleaner option that this? Seems like something flex should be able to handle no?

rohitpaulk (changes requested):
Added note re: usage of padding/absolute positioning

app/components/course-page/comment-timeline-item.hbs
VasylMarchuk: This is going to be tricky to explain (because it's CSS, duh), be I have to start backwards:

#### It all starts with the `white-space: pre` property on the `code` element, which makes it want to render to full maximum width allowed to it:

Знімок екрана 2025-04-11 о 10 49 33

#### In turn, maximum allowed width is determined by this `width: 100%` property:

Знімок екрана 2025-04-11 о 10 50 48

#### And in turn, in the context of that element, 100% width is the width of it's parent:

Знімок екрана 2025-04-11 о 10 52 47

#### And so:

Because the `div` with the comment card (the one with `w-full`) already "occupies" 100% of the parent, there is no place remaining for the avatar, and it squeezes the div out of the layout.

I have spent way too much time on this and I did try all sorts of magic with `flex`, shrinking, growing etc, but in the end it doesn't seem like flex is the right tool here (all because the exact size of that `code` is unspecified, but it wants to both occupy 100% of _something_, and keep the content one one line without wrapping).

Grid might be a better replacement, but seems like too much effort, and also, this `position: absolute` is not that bad, and actually seems to make sense together with that `w-full`.

@rohitpaulk

rohitpaulk: Note added in slack re: trying `min-width-0`

rohitpaulk (approved):

PR #2784, opened 2025-04-06
Pre-render Contests using Vercel Prerender Functions & remove middleware

VasylMarchuk:
Related to #2607

### Brief

This adds pre-rendering of Contests using Prerender Functions, and removes the middleware that's currently used to inject meta tags. Meta tags for Contests are now properly handled in their ember route.

### Details

- Added Prerender function `/contests/[contest].func`
- Handle `/contests/:contest` Vercel route using the above function
- Added proper meta tags handling to `ContestRoute` in Ember
- Pre-render `/contests/weekly-7` with Prember (to catch potential errors at deployment)
- Remove no longer used `app/utils/replace-meta-tag.js`

### Checklist


app/routes/contest.ts
```diff
@@ -1,72 +1,132 @@
...
-    const languages = (await this.store.findAll('language')) as unknown as LanguageModel[];
-
-    return {
-      contest,
-      allContests,
-      topLeaderboardEntries,
-      surroundingLeaderboardEntries,
-      languages,
-    } as ModelType;
   }
```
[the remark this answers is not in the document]
VasylMarchuk: It has a default value of `[]` in the controller itself

app/components/contest-page/leaderboard-card.ts
```diff
@@ -20,27 +20,31 @@ interface Signature {
 export default class ContestPageLeaderboardCardComponent extends Component<Signature> {
   @service declare authenticator: AuthenticatorService;
 
+  get filteredTopEntries() {
```
rohitpaulk, line 23: @VasylMarchuk in what case would `topEntries` contain an entry that doesn't belong to this leaderboard?

VasylMarchuk: Because `topEntries` returned by the route is the result of `findAll`, which returns **all entries** currently in the store, **unfiltered**. The result of `findAll` is also a magic `Proxy` which can be updated later (background refresh), glimmer understands this update and re-triggers the filtering getter. I can move the filtering getter from the component into the controller if you'd like, so that the component receives already filtered topEntries.

But notice that all other things returned from the model hook (languages, allContests) are also unfiltered.

VasylMarchuk: In other words, there's a difference:

```ts
model () {
  return this.store.findAll(...); // <-- RecordArray, background refresh handled by Glimmer
  return this.store.findAll(...).filter(); // <-- Array, background refresh will not be reflected by the template
}
```

So the model hook ideally needs to return `RecordArray`s, plain results of `findAll`, which then need to be filtered in controller/component.

VasylMarchuk: I've moved `filteredTopEntries` from `LeaderboardCardComponent` into `ContestController` (renamed as `filteredTopLeaderboardEntries`)

rohitpaulk (approved):

PR #2788, opened 2025-04-07
Greyed out released stuff and try adding a test(?)

Arpan-206:
**Checklist**:

## Summary by CodeRabbit

- **New Features**
  - Course extension and idea cards now provide clearer visual feedback by displaying with reduced opacity when marked as released.
  
- **Tests**
  - Enhanced tests to verify that the visual indication (greyed-out style) is applied correctly to the cards for anonymous users, ensuring accurate representation of their development status.
  - Added assertions to check development status labels and visual states for released and not started course ideas.

### DIFF
--- a/mirage/utils/create-course-extension-ideas.js
+++ b/mirage/utils/create-course-extension-ideas.js
@@ -9,6 +9,7 @@ export default function createCourseExtensionIdeas(server) {
       slug: courseExtensionIdeaData.slug,
       descriptionMarkdown: courseExtensionIdeaData.description_md,
       votesCount: 0,
+      developmentStatus: courseExtensionIdeaData.development_status,
     });
   }
 }
--- a/mirage/utils/create-course-extension-ideas.js
+++ b/mirage/utils/create-course-extension-ideas.js
@@ -7,7 +7,7 @@ export default function createCourseExtensionIdeas(server) {
       createdAt: new Date(),
       descriptionMarkdown: courseExtensionIdeaData.description_md,
       developmentStatus: courseExtensionIdeaData.development_status,
-      developmentStatusLabelText: courseExtensionIdeaData.development_status_label,
+      developmentStatusLabelText: courseExtensionIdeaData.development_status,
       name: courseExtensionIdeaData.name,
       slug: courseExtensionIdeaData.slug,
       votesCount: 0,
--- a/mirage/utils/create-course-extension-ideas.js
+++ b/mirage/utils/create-course-extension-ideas.js
@@ -3,13 +3,13 @@ import courseExtensionIdeasData from '../data/course-extension-ideas';
 export default function createCourseExtensionIdeas(server) {
   for (const courseExtensionIdeaData of courseExtensionIdeasData) {
     server.create('course-extension-idea', {
-      createdAt: new Date(),
       course: server.schema.courses.findBy({ slug: courseExtensionIdeaData.course_slug }),
+      createdAt: new Date(),
+      descriptionMarkdown: courseExtensionIdeaData.description_md,
+      developmentStatus: courseExtensionIdeaData.development_status,
       name: courseExtensionIdeaData.name,
       slug: courseExtensionIdeaData.slug,
-      descriptionMarkdown: courseExtensionIdeaData.description_md,
       votesCount: 0,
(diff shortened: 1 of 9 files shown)


tests/pages/vote-page.js
```diff
@@ -7,13 +7,15 @@ export default createPage({
     hoverOnDevelopmentStatusLabel: triggerable('mouseenter', '[data-test-development-status-label]'),
     name: text('[data-test-course-extension-idea-name]'),
     voteButtonText: text('[data-test-vote-button]'),
+    isGreyedOut: (card) => card.hasClass('opacity-50'),
```
rohitpaulk, line 10: When there's no clear order or order doesn't matter, we sort things alphabetically - so this would appear after `hoverOn...`. This helps avoid that extra decision around where to place a line, also helps keep things more consistent across the codebase

rohitpaulk (approved):
Good to merge, I've left a loom in slack re: how to add tests

tests/pages/vote-page.js
Arpan-206: Addressed in later commits!

mirage/utils/create-course-extension-ideas.js
```diff
@@ -9,6 +9,7 @@ export default function createCourseExtensionIdeas(server) {
       slug: courseExtensionIdeaData.slug,
       descriptionMarkdown: courseExtensionIdeaData.description_md,
       votesCount: 0,
+      developmentStatus: courseExtensionIdeaData.development_status,
```
rohitpaulk, line 12: Looks like the existing code didn't exactly follow this, but the properties here should be alphabetically ordered too

mirage/utils/create-course-ideas.js
```diff
@@ -7,6 +7,7 @@ export default function createCourseIdeas(server) {
       name: courseIdeaData.name,
       descriptionMarkdown: courseIdeaData.description_md,
       votesCount: 0,
+      developmentStatus: courseIdeaData.development_status,
```
rohitpaulk, line 10: Same note as above

tests/acceptance/vote-page/course-extension-ideas-test.js
```diff
@@ -25,6 +25,7 @@ module('Acceptance | vote-page | course-extension-ideas', function (hooks) {
     assert.strictEqual(votePage.selectedCourseName, 'Build your own Redis');
     assert.strictEqual(votePage.courseExtensionIdeaCards.length, 2);
     assert.strictEqual(votePage.findCourseExtensionIdeaCard(courseExtensionIdea.name).voteButtonText, '1 vote');
+    assert.true(votePage.findCourseExtensionIdeaCard(courseExtensionIdea.name).isGreyedOut);
```
rohitpaulk, line 28: It's not super clear why this should be grayed out + we aren't testing whether other ideas are not grayed out. A more easy-to-read test would be something like this: 

```
const releasedIdeaCard = votePage.findCourseExtensionIdeaCard(...)
const notStartedIdeaCard = votePage.findCourseExtensionIdeaCard(...)

assert.strictEqual(releasedIdeaCard.developmentStatusLabelText, "released")
assert.true(releasedIdeaCard.isGreyedOut)

assert.false(releasedIdeaCard.developmentStatusLabel.isPresent)
assert.false(releasedIdeaCard.isGreyedOut)
```

This mimics what a human would look for - are released ideas grayed out and marked as so? And are the other ones not greyed out?

rohitpaulk (approved):
Minor suggestions, good to merge once addressed!

mirage/utils/create-course-extension-ideas.js
Arpan-206: do you want me to re-order all, I did think of it while writing the code but since existing wasn't following it

tests/acceptance/vote-page/course-extension-ideas-test.js
Arpan-206: makes sense

mirage/utils/create-course-extension-ideas.js
rohitpaulk: Yep let's reorder -- order doesn't matter here so let's correct the previous mistake

PR #2805, opened 2025-04-17
Shortened Header Text + New URL + Set Up Redirect

Arpan-206:
**Checklist**:

## Summary by CodeRabbit

- **New Features**
	- Added permanent redirects so that visits to old vote page URLs are automatically sent to the new challenge and extension paths.
- **Bug Fixes**
	- Updated tab labels on the vote page for improved clarity.
	- Updated vote page URLs and navigation to reflect new challenge and extension naming.
- **Tests**
	- Updated test URLs and comments to match new vote page paths.

### DIFF
--- a/app/router.ts
+++ b/app/router.ts
@@ -94,8 +94,8 @@ Router.map(function () {
 
   // TODO: Add dark mode support
   this.route('vote', function () {
-    this.route('course-ideas', { path: '/challenge-ideas' });
-    this.route('course-extension-ideas', { path: '/challenge-extension-ideas' });
+    this.route('course-ideas', { path: '/challenges' });
+    this.route('course-extension-ideas', { path: '/challenge-extensions' });
   });
 
   this.route('welcome');
--- a/app/components/vote-page/header.hbs
+++ b/app/components/vote-page/header.hbs
@@ -5,6 +5,6 @@
 </p>
 
 <div class="flex mb-8 border-b overflow-x-auto">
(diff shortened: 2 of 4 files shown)


app/router.ts
```diff
@@ -94,8 +94,11 @@ Router.map(function () {
 
   // TODO: Add dark mode support
   this.route('vote', function () {
-    this.route('course-ideas', { path: '/challenge-ideas' });
-    this.route('course-extension-ideas', { path: '/challenge-extension-ideas' });
+    this.route('course-ideas', { path: '/challenges' });
+    this.route('course-extension-ideas', { path: '/challenge-extensions' });
+    // Add redirects for old paths
+    this.route('redirect-challenge-ideas', { path: '/challenge-ideas' });
```
rohitpaulk, line 100: @Arpan-206 this is easier to place in [vercel.json](https://github.com/codecrafters-io/frontend/blob/main/vercel.json), doesn't need a separate route definition :)

app/routes/vote/challenge-extension-ideas.js
```diff
@@ -0,0 +1,8 @@
+import Route from '@ember/routing/route';
+
+export default class VoteChallengeExtensionIdeasRoute extends Route {
```
rohitpaulk, line 3: @Arpan-206 why do we need these files? The route name is `course-extension-ideas`, and we already have a route file for it. This seems like it's not used at all?

rohitpaulk (approved):
Good to merge once notes are addressed -- we shouldn't need to change any route files, just override `path` in `router.ts`. We can also move the redirect to vercel. We can't add tests for the vercel-level redirect but it's pretty straightforward functionality so okay to not test

PR #2807, opened 2025-04-20
Animation on hover w/ Gleam Logo

Arpan-206:
**Checklist**:

## Summary by CodeRabbit

- **New Features**
  - Updated dependencies to include the Rive animation library.
  - Enhanced track page header with a conditional animated Gleam logo.
  - Introduced a new `RiveAnimation` component for rendering animations on a canvas with automatic resizing and error handling.
  - Added a dedicated demo page showcasing various Rive animations with different sizes and files.
  - Included a new route and navigation tab for the Rive animation demo.
  - Created a new template demonstrating multiple Rive animations with different configurations.
- **Tests**
  - Added integration tests for the `RiveAnimation` component, verifying rendering and animation display with different sources.


Arpan-206:
@rohitpaulk  would love an opinion on testing, required or not.

Arpan-206:
https://www.loom.com/share/85e5c7a87b3c4f03897153e988134f3d?sid=6f78cd08-a4d6-4e4b-9899-a5c27653a10a

for a preview of the animation

app/components/gleam-logo.hbs
```diff
@@ -0,0 +1,9 @@
+<div
+  class="gleam-logo-container"
+  {{did-insert this.setupRive}}
```
rohitpaulk, line 3: Our convention for handlers is to always name them after the event (i.e. something like  `this.handleDidInsert` in this case). Let's us now have to worry about naming each of these + allows us to easily edit handlers to add more functionality without having to rename them.

app/components/gleam-logo.ts
```diff
@@ -0,0 +1,130 @@
+import Component from '@glimmer/component';
+import { tracked } from '@glimmer/tracking';
+import { action } from '@ember/object';
+import { Rive } from '@rive-app/canvas';
+
+interface GleamLogoSignature {
+  Element: HTMLDivElement;
+  Args: {
+    class?: string;
```
rohitpaulk, line 9: @Arpan-206 in Ember you don't need to accept DOM-level attributes like `class`/`style`, these just get passed along with `...attributes`. You only need to use `Args` for things that you'll explicitly pass in

app/components/gleam-logo.ts
```diff
@@ -0,0 +1,130 @@
...
+    style?: string;
+    [key: string]: unknown;
```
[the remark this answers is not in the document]
Arpan-206: Screenshot 2025-04-21 at 18 59 57

due to this

app/components/language-logo.ts
```diff
@@ -15,5 +15,6 @@ export default class LanguageLogoComponent extends Component<Signature> {}
 declare module '@glint/environment-ember-loose/registry' {
   export default interface Registry {
     LanguageLogo: typeof LanguageLogoComponent;
+    GleamLogo: typeof import('./gleam-logo').default;
```
rohitpaulk, line 18: @Arpan-206 why is this registered here and not in `gleam-logo.ts` (like every other component is)?

app/components/language-logo.hbs
```diff
@@ -1,4 +1,6 @@
-{{#if (eq @variant "color")}}
+{{#if (and (eq @language.slug "gleam") (eq @variant "color"))}}
+  <GleamLogo ...attributes />
```
rohitpaulk, line 2: @Arpan-206 let's not assume that every color language logo usage of Gleam is going to be animated for now - let's keep this explicit (use `GleamLogo` instead of `LanguageLogo` in the tracks template). 

You can imagine using a color logo in things like user profile pages where we might not want this animation to occur

app/components/gleam-logo.ts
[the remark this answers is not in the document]
Arpan-206: well the first one, forgot to remove the signature

app/components/gleam-logo.ts
```diff
@@ -0,0 +1,130 @@
...
+
+  @action
+  setupRive(element: HTMLDivElement) {
+    this.container = element;
+
+    try {
+      const canvas = document.createElement('canvas');
+
+      // Set the canvas size to be 4x the original size for better quality
+      canvas.width = 141 * 4;
```
rohitpaulk, line 66: @Arpan-206 where do these numbers like 141 and 144 come from? Our goal is to be able to use Rive animations as a "component" in the future, without having to hardcode values specific to each animation file. Is there a way to do that here? Something like saying we want the height to fit the container and the width to just adjust accordingly? Similar to how we handle SVG images

app/components/gleam-logo.ts
```diff
@@ -0,0 +1,130 @@
...
+      canvas.height = 144 * 4;
+
+      // Scale the canvas to fit the container while maintaining aspect ratio
+      canvas.style.width = '100%';
+      canvas.style.height = '100%';
```
rohitpaulk, line 71: @Arpan-206 let's confirm if this actually does preserve aspect ratio? According to the numbers above the file isn't exactly square shaped

app/components/gleam-logo.ts
```diff
@@ -0,0 +1,130 @@
...
+
+      element.appendChild(canvas);
+
+      this.riveInstance = new Rive({
+        src: '/assets/animations/gleam_logo_animation.riv',
+        canvas: canvas,
+        autoplay: true,
+        automaticallyHandleEvents: true,
+        onLoad: () => {
+          // Set up interval to play animation every 15 seconds
```
rohitpaulk, line 82: We don't need this to play every 15 seconds @Arpan-206 - the spec only says on load, and on hover

Arpan-206: oh that was for debugging, removed now, thanks!

app/components/gleam-logo.ts
```diff
@@ -0,0 +1,130 @@
...
+              }
+            }
+
+            // Simulate mouse leave after a short delay
+            setTimeout(() => {
+              this.handleMouseLeave();
+            }, 2000);
+          }, 15000);
+          // Play initial animation with State Machine 2
+          setTimeout(() => {
```
rohitpaulk, line 103: Does `autoplay` not cover this? And it not, what does autoplay do?

app/components/gleam-logo.ts
```diff
@@ -0,0 +1,130 @@
...
+                    }
+                  }, 800);
+                }
+              }
+            }
+          }, 500);
+        },
+      });
+    } catch (error: unknown) {
+      console.error('Error setting up Rive:', error);
```
rohitpaulk, line 127: Let's report this to sentry, don't just swallow and ignore

app/styles/components/gleam-logo.css
```diff
@@ -0,0 +1,12 @@
+.gleam-logo-container {
```
rohitpaulk, line 1: @Arpan-206 all of this can be done via tailwind, you don't need specific styles for this

app/styles/components/gleam-logo.css
```diff
@@ -0,0 +1,12 @@
...
+  display: block;
+  width: 100%;
+  height: 100%;
+}
+
+.gleam-logo-container canvas {
+  display: block;
+  width: 100%;
+  height: 100%;
+  object-fit: contain;
```
rohitpaulk, line 11: We're setting the same styles via JS, why are we duplicating here?

Arpan-206: I really should learn to remove my debugs better sorry!

app/components/gleam-logo.hbs
Arpan-206: noted!

rohitpaulk (changes requested):
Added one round of comments, this needs to be FAR simpler & clean - too complicated atm. Will take another look once the first set of comments are addressed, also asked Arief about how we can handle the reverse animation

app/components/gleam-logo.ts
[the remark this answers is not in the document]
rohitpaulk: We don't have to do this anywhere else in the repository, try figuring out why?

app/styles/components/gleam-logo.css
Arpan-206: right!

app/components/gleam-logo.ts
```diff
@@ -0,0 +1,114 @@
...
+import Component from '@glimmer/component';
+import { tracked } from '@glimmer/tracking';
+import { Fit, Layout, Rive } from '@rive-app/canvas';
+import * as Sentry from '@sentry/ember';
+
+interface GleamLogoSignature {
+  Element: HTMLDivElement;
+
+  Args: {
+    height: number;
```
rohitpaulk, line 11: Let's try to avoid needing a number to be passed in and see if we can just have it fit the container / parent element

app/components/gleam-logo.ts
```diff
@@ -0,0 +1,114 @@
...
+}
+
+export default class GleamLogoComponent extends Component<GleamLogoSignature> {
+  animationInterval: number | null = null;
+  container: HTMLElement | null = null;
+  @tracked riveInstance: Rive | null = null;
+
+  get containerStyle(): string {
+    // Ensure minimum size on mobile while maintaining aspect ratio
+    const minSize = Math.min(this.args.height, 200);
```
rohitpaulk, line 26: @Arpan-206 where does the "200" number come from?

app/components/gleam-logo.ts
```diff
@@ -0,0 +1,114 @@
...
+
+  @action
+  handleDidInsert(element: HTMLDivElement) {
+    this.container = element;
+
+    try {
+      const canvas = document.createElement('canvas');
+
+      // Set initial canvas size for high quality
+      const baseSize = 400; // Base size for quality
```
rohitpaulk, line 52: @Arpan-206 let's see if we can avoid this magic number too and just have it pick a sensible value (like based on pixel density & container height)

app/components/gleam-logo.ts
```diff
@@ -0,0 +1,114 @@
...
+      canvas.style.height = '100%';
+      canvas.style.display = 'block';
+      canvas.style.maxWidth = '100%'; // Ensure it doesn't overflow on mobile
+
+      element.appendChild(canvas);
+
+      this.riveInstance = new Rive({
+        src: '/assets/animations/gleam_logo_animation.riv',
+        canvas: canvas,
+        stateMachines: 'State Machine 1',
```
rohitpaulk, line 67: It is mandatory to pass in state machine names? Will this work without it?

app/components/gleam-logo.ts
```diff
@@ -0,0 +1,114 @@
...
+        }),
+        onLoad: () => {
+          if (this.riveInstance) {
+            // Initial resize
+            this.riveInstance.resizeDrawingSurfaceToCanvas();
+
+            const inputs = this.riveInstance.stateMachineInputs('State Machine 1');
+
+            // Try to trigger hover state
+            if (inputs) {
```
rohitpaulk, line 81: Let's move most logic to the animatino file itself, and maybe name the animation / state machine `onLoad`

app/components/language-logo.hbs
```diff
@@ -1,4 +1,6 @@
-{{#if (eq @variant "color")}}
+{{#if (and (eq @language.slug "gleam") (eq @variant "color"))}}
+  <GleamLogo @height={{144}} ...attributes />
```
rohitpaulk, line 2: @Arpan-206 I mentioned this before, let's move this out to the tracks page

Screenshot 2025-04-30 at 12 29 06

rohitpaulk (changes requested):
Notes added

app/components/gleam-logo.ts
```diff
@@ -0,0 +1,134 @@
...
+      const canvas = document.createElement('canvas');
+
+      // Calculate base size based on device pixel ratio and container size
+      const pixelRatio = window.devicePixelRatio || 1;
+      const containerSize = this.args.size || this.containerSize;
+      // Use 2x the container size for high quality, but cap at 4x for performance
+      const qualityMultiplier = Math.min(4, Math.max(2, pixelRatio));
+      const baseSize = Math.round(containerSize * qualityMultiplier);
+
+      canvas.width = baseSize;
```
rohitpaulk, line 73: Here are we assuming that all animations are square? Don't think we can do this, very possible to have animations that aren't square (even some of our language logos aren't square)

app/components/gleam-logo.ts
```diff
@@ -0,0 +1,134 @@
...
+      canvas.height = baseSize;
+
+      // Let the canvas scale naturally within its container
+      canvas.style.width = '100%';
```
rohitpaulk, line 77: @Arpan-206 is there a reason for us to create this canvas element "manually" annd `appendChild` rather than just place a `canvas` inside the template, and use `didInsert` on that instead? That way you can even use tailwind styles to avoid the manual setting of width, display etc.

app/components/gleam-logo.ts
```diff
@@ -0,0 +1,134 @@
...
+      this.riveInstance = new Rive({
+        src: '/assets/animations/gleam_logo_animation.riv',
+        canvas: canvas,
+        stateMachines: 'Default',
+        autoplay: true,
+        automaticallyHandleEvents: true,
+        layout: new Layout({
+          fit: Fit.Contain,
+        }),
+        onLoad: async () => {
```
rohitpaulk, line 93: So the "onLoad" state machine is gone now? `Default` handles both events and the autoplay?

app/components/gleam-logo.ts
```diff
@@ -0,0 +1,134 @@
...
+          if (this.riveInstance) {
+            // Initial resize
+            this.riveInstance.resizeDrawingSurfaceToCanvas();
```
rohitpaulk, line 96:
```suggestion
            this.riveInstance?.resizeDrawingSurfaceToCanvas();
```

(Shorthand to replace the if check)

app/components/gleam-logo.ts
```diff
@@ -0,0 +1,134 @@
...
+          }
+        },
+      });
+    } catch (error: unknown) {
+      console.error('Error setting up Rive:', error);
+      Sentry.captureException(error, {
```
rohitpaulk, line 102: Just `captureException` without extra tags should be fine @Arpan-206 - sentry should automatically capture context like component etc. Example: https://codecrafters.sentry.io/issues/6269915267/?project=5922961&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=issue-stream&stream_index=5

app/components/language-logo.hbs
rohitpaulk: @Arpan-206 this hasn't been addressed

app/components/gleam-logo.ts
rohitpaulk: @Arpan-206 to ensure that we aren't doing anything specific to Gleam, let's try this with a few different Rive animations (download random ones off the internet). Extract a `RiveAnimation` component and add a tab for it with some examples to https://app.codecrafters.io/demo. Try to get varied kinds - with/without transparency, square vs. rectangle etc.

tests/integration/components/gleam-logo-test.js
```diff
@@ -0,0 +1,134 @@
...
+      }
+    };
+  });
+
+  hooks.afterEach(function () {
+    window.Rive = this.originalRive;
+  });
+
+  test('it renders and initializes correctly', async function (assert) {
+    await render(hbs`<GleamLogo @height={{200}} />`);
```
rohitpaulk, line 87: We're passing in `height` here, but the argument is called `size` in the component file

rohitpaulk (changes requested):
Comments added

app/components/gleam-logo.ts
Arpan-206: yes, we're using layers as their support said

Arpan-206:
https://frontend-j2hjvwpux.ccio.dev/demo/rive-animation

Here's a demo!

rohitpaulk:
@Arpan-206 in the demo these images get squished and lose aspect ratio: 


https://github.com/user-attachments/assets/701e0581-457e-4560-a1f5-bf055877f266

rohitpaulk (changes requested):
Added note re: maintaining aspect ratio

rohitpaulk:
@Arpan-206 resize doesn't seem to be fixed: 

https://github.com/user-attachments/assets/999777fd-ed6f-475d-9dbb-049eef582392

rohitpaulk (changes requested):
(resize issue persists)

Arpan-206:
https://frontend-nx6125hwh.ccio.dev/demo/rive-animation

rohitpaulk:
@Arpan-206 this doesn't seem to work on mobile, can you check why please? 

https://github.com/user-attachments/assets/8b98af33-3ce4-4443-9d09-2b6b80ff8374

app/components/rive-animation.hbs
```diff
@@ -0,0 +1,3 @@
+<div {{did-insert this.handleDidInsert}} {{will-destroy this.handleWillDestroy}} style={{this.containerStyle}} ...attributes>
+  <canvas class="w-full h-full block max-w-full"></canvas>
```
rohitpaulk, line 2: See if we can use canvas directly with no surrounding element

app/components/rive-animation.ts
```diff
@@ -0,0 +1,157 @@
...
+  };
+}
+
+export default class RiveAnimationComponent extends Component<RiveAnimationSignature> {
+  animationInterval: number | null = null;
+  container: HTMLElement | null = null;
+  resizeObserver: ResizeObserver | null = null;
+  @tracked riveInstance: Rive | null = null;
+  @tracked containerWidth: number = 0;
+  @tracked containerHeight: number = 0;
```
rohitpaulk, line 20: See if we can avoid tracking `containerHeight` / `containerWidth` - and just rely on natural HTML sizing

app/components/rive-animation.ts
```diff
@@ -0,0 +1,157 @@
...
+  }
+
+  private updateDimensions(width: number, height: number) {
+    if (!this.animationAspectRatio) return;
+
+    const containerAspectRatio = width / height;
+    let newWidth = width;
+    let newHeight = height;
+
+    if (containerAspectRatio > this.animationAspectRatio) {
```
rohitpaulk, line 128: See if we can avoid all this manual calculation by just relying on Rive's "fit" layout implementation, or by using native HTML scaling

app/components/rive-animation.ts
```diff
@@ -0,0 +1,81 @@
+import { action } from '@ember/object';
+import Component from '@glimmer/component';
+import { tracked } from '@glimmer/tracking';
+import { Fit, Layout, Rive } from '@rive-app/canvas';
+import * as Sentry from '@sentry/ember';
+
+interface RiveAnimationSignature {
```
rohitpaulk, line 7:
```suggestion
interface Signature {
```

(Consistent with our other component files)

app/components/rive-animation.ts
```diff
@@ -0,0 +1,81 @@
...
+        }),
+        onLoad: async () => {
+          if (this.riveInstance) {
+            // Get the animation bounds and calculate aspect ratio
+            const bounds = this.riveInstance.bounds;
+
+            if (bounds) {
+              const width = bounds.maxX - bounds.minX;
+              const height = bounds.maxY - bounds.minY;
+              this.animationAspectRatio = width / height;
```
rohitpaulk, line 49: Where is `animationAspectRatio` used?

tests/integration/components/rive-animation-test.js
```diff
@@ -0,0 +1,176 @@
...
+    assert.ok(inputs, 'State machine inputs exist');
+
+    const hoverInput = inputs.find((input) => input.name.toLowerCase().includes('hover'));
+    assert.ok(hoverInput, 'Hover input exists');
+
+    // Verify initial hover state
+    assert.false(hoverInput.value, 'Hover input is initially false');
+
+    // Wait for hover out timeout
+    assert.false(hoverInput.value, 'Hover input is set to false after timeout');
```
rohitpaulk, line 148: Where is the "timeout" here? Also seems like we're just testing the mocked static array above here, don't really see the point of the test

tests/integration/components/rive-animation-test.js
```diff
@@ -0,0 +1,176 @@
...
+  });
+
+  test('it works with different animation files', async function (assert) {
+    await render(hbs`
+      <div class="w-[200px] h-[200px]">
+        <RiveAnimation @src="/assets/animations/pig_cuddly.riv" />
+      </div>
+    `);
+
+    const container = this.element.querySelector('div > div');
```
rohitpaulk, line 158: I thought we're using canvas and not `div` for the component, so how should this be passing? Maybe it's catching other `div` elements?

tests/integration/components/rive-animation-test.js
```diff
@@ -0,0 +1,176 @@
+import { module, test } from 'qunit';
+import { setupRenderingTest } from 'codecrafters-frontend/tests/helpers';
+import { render } from '@ember/test-helpers';
+import { hbs } from 'ember-cli-htmlbars';
+
+// Mock Rive class for testing
+class MockRive {
```
rohitpaulk, line 7: In general this seems to be a pretty clumsy testing strategy - we don't get a lot of benefit out of it. I'd recommend dropping this mocking entirely and just checking that (a) there are no errors and (b) some kind of assertion that the canvas has things drawn inside it (i.e. an "end-to-end" test)

rohitpaulk (changes requested):
Notes added!

tests/integration/components/rive-animation-test.js
```diff
@@ -0,0 +1,83 @@
...
+
+    // Check that canvas is rendered with correct classes
+    const canvas = this.element.querySelector('canvas');
+    assert.ok(canvas, 'Canvas element exists');
+    assert.strictEqual(canvas?.className, 'w-full h-full block max-w-full', 'Canvas has correct classes');
+
+    // Wait for the animation to load and render
+    await settled();
+    // Give Rive time to actually render
+    await new Promise((resolve) => setTimeout(resolve, 200));
```
rohitpaulk, line 24: Use `waitUntil` from ember test helper in this case, makes tests less flaky

PR #2825, opened 2025-05-12
Add Code example insights index page

ryan-gang:
**Checklist**:

## Summary by CodeRabbit

- **New Features**
  - Introduced a "Code Examples" tab in the course admin interface, providing access to a new insights page.
  - Added a code example insights index page to view and filter solution statistics by programming language for each course stage.
  - Implemented detailed stage-level insights showing solution counts, upvotes, downvotes, and median changed lines with color-coded visual indicators and explanatory tooltips.
  - Enabled language selection with dynamic filtering and UI feedback for selected or missing languages.
  - Enhanced navigation with clickable stage items opening detailed insights in a new tab.
  - Added comprehensive acceptance tests to ensure correct language defaulting and stage list rendering.

- **Chores**
  - Updated .gitignore to exclude additional files and directories.


ryan-gang:
Don't want this to be delayed. 
So opening this for review early. (We can wait and see how it looks with solid data before deploying.)

ryan-gang:
As this is an admin panel page, didn't write tests too. Lemme know if I should add some.

app/templates/course-admin/code-example-insights-index.hbs
```diff
@@ -0,0 +1,70 @@
...
+      <div>
+        <h3 class="text-lg font-bold text-gray-800 mb-1">
+          Code Examples 🚧 (Work in progress!)
+        </h3>
+        <div class="text-xs text-gray-400">
+          Summaries of code examples submitted by users.
+        </div>
+      </div>
+      <div>
+        {{#if @model.selectedLanguage}}
```
rohitpaulk, line 13: Let's remove this element/flow - since we already have the dropdown, users can use that to toggle language. We don't need to allow them to "clear" the selection, a user would always just want to toggle the language

app/controllers/course-admin/code-example-insights-index.ts
```diff
@@ -0,0 +1,27 @@
...
+import type { ModelType } from 'codecrafters-frontend/routes/course-admin/code-example-insights-index';
+
+export default class CodeExampleInsightsIndexController extends Controller {
+  @service declare router: RouterService;
+  @service declare store: Store;
+
+  declare model: ModelType;
+
+  @action
+  onLanguageChange(language: LanguageModel | null) {
```
rohitpaulk, line 16:
```suggestion
  onLanguageChange(language: LanguageModel) {
```

Let's not allowing transitioning to a state where a language is not present

rohitpaulk (approved):

rohitpaulk:
> As this is an admin panel page, didn't write tests. Lemme know if I should add some.

We should write tests for this since it's the kind that will be exposed to course authors. The "admin-only" wording means things like the demo page, debug page etc. Course author is a bit of a gray area - it's only us staff for now but we will have external authors in the future

PR #2917, opened 2025-06-25
Change Roadmap Idea Ordering

Arpan-206:
**Checklist**:

### DIFF
--- a/app/models/course-extension-idea.ts
+++ b/app/models/course-extension-idea.ts
@@ -43,7 +43,11 @@ export default class CourseExtensionIdeaModel extends Model {
       released: 1,
     }[this.developmentStatus];
 
-    return `${reverseSortPositionFromDevelopmentStatus}-${this.createdAt.toISOString()}`;
+    // Use vote count for secondary sorting (higher votes = higher priority)
+    // Pad vote count with zeros to ensure proper string sorting
+    const paddedVoteCount = this.votesCount.toString().padStart(10, '0');
+
+    return `${reverseSortPositionFromDevelopmentStatus}-${paddedVoteCount}`;
   }
 
   async vote() {
--- a/app/models/course-idea.ts
+++ b/app/models/course-idea.ts
@@ -35,7 +35,11 @@ export default class CourseIdeaModel extends Model {
       released: 1,
     }[this.developmentStatus];
 
-    return `${reverseSortPositionFromDevelopmentStatus}-${this.createdAt.toISOString()}`;
+    // Use vote count for secondary sorting (higher votes = higher priority)
+    // Pad vote count with zeros to ensure proper string sorting
+    const paddedVoteCount = this.votesCount.toString().padStart(10, '0');
+
+    return `${reverseSortPositionFromDevelopmentStatus}-${paddedVoteCount}`;
   }
 
   async vote(): Promise<CourseIdeaVoteModel> {


rohitpaulk (changes requested):
Overall the util file feels kinda complicated for what it does, let's try to simplify. Added some notes on the file, other ideas: 

- Try to remove the "reverse" idea and just do `getSortPositionForRoadmapPage` - reverse was needed previously because we were sorting by created date, we aren't doing that anymore. 
- Try to compute each key separately into an array, and then join them together at the end in case that simplifies the logic.

app/utils/roadmap-sorting.ts
```diff
@@ -0,0 +1,53 @@
...
+      not_started: 2,
+      released: 3,
+    }[developmentStatus] ?? 4; // Default to lowest priority if status is unknown
+
+  const sortKeys: string[] = [statusPriority.toString()];
+
+  if (currentUserId) {
+    // For authenticated users: fixed order per user per week
+    const weekSeed = getRoughWeekSeed(date);
+    const userWeekSeed = `${currentUserId}-${weekSeed}`;
```
rohitpaulk, line 29: It this is only used in the second branch (`not_started`), instantiate it there instead - placing it outside the branches makes it look like this is common to 1+ branches.

app/utils/roadmap-sorting.ts
```diff
@@ -0,0 +1,53 @@
...
+    } else if (developmentStatus === 'not_started') {
+      // Not started: sort by random but fixed per user per week
+      const rng = seedrandom(`${userWeekSeed}-${id}`);
+      const randomSeed = Math.floor(rng() * 1000000); // Generate a number between 0-999999
+      const paddedRandomSeed = randomSeed.toString().padStart(10, '0');
+      sortKeys.push(paddedRandomSeed);
+    } else {
+      // Released: sort by vote count descending (highest votes first)
+      const invertedVoteCount = (999999999 - votesCount).toString().padStart(10, '0');
+      sortKeys.push(invertedVoteCount);
```
rohitpaulk, line 44: This is the same as the first branch - would group together to simplify, something like this: 

```
if (status === "not_started") {
   // Not started: sort by random but fixed per user per week
   ...
} else {
  // Sort other ideas by vote count descending
}
```

app/utils/roadmap-sorting.ts
```diff
@@ -0,0 +1,53 @@
...
+  const sortKeys: string[] = [statusPriority.toString()];
+
+  if (currentUserId) {
+    // For authenticated users: fixed order per user per week
+    const weekSeed = getRoughWeekSeed(date);
+    const userWeekSeed = `${currentUserId}-${weekSeed}`;
+
+    if (developmentStatus === 'in_progress') {
+      // In progress: sort by vote count descending (highest votes first)
+      const invertedVoteCount = (999999999 - votesCount).toString().padStart(10, '0');
```
rohitpaulk, line 33:
```suggestion
      const invertedVoteCountKey = (999999999 - votesCount).toString().padStart(10, '0');
```

(or even `invertedVoteCountSortKey`)

The variable name might suggest that the value is an int, adding `key` to it makes it clear that it's a string to be used inside `sortKeys`

app/utils/roadmap-sorting.ts
rohitpaulk: Looking at all the branches in this function as a whole it kinda looks like the only special treatment is if a user is logged in, and the idea is "not_started". So could simplify further: 

```
if (currentUserId && status === "not_started") {
  ...
} else {
  ... 
}
```

rohitpaulk: The branches are what make this hard to read, without those I think this'll be simpler

app/utils/roadmap-sorting.ts
```diff
@@ -0,0 +1,53 @@
...
+}
+
+export function getSortPositionForRoadmapPage(
+  developmentStatus: string,
+  votesCount: number,
+  id: string,
+  date: number,
+  currentUserId?: string | null,
+): string {
+  const statusPriority =
```
rohitpaulk, line 17:
```suggestion
  const statusSortKey =
```

(We can make this a string right away, no need to call `.toString()`, and adding `sortKey` to the name makes it clear that it's used in `sortKeys` below and avoids the additional "priority" word)

app/utils/roadmap-sorting.ts
```diff
@@ -0,0 +1,53 @@
+import seedrandom from 'seedrandom';
+
+export function getRoughWeekSeed(date: number): number {
+  const now = date;
```
rohitpaulk, line 4: Pointless indirection - just do `return date - (date % weekMs)`.

rohitpaulk (approved):

PR #2928, opened 2025-06-28
Redrawing the roadmap (vote) pages

Arpan-206:
**Checklist**:


Arpan-206:
My bad thought I'd converted to draft but clearly not.

PR #2929, opened 2025-06-30
Using a different technique to gray out logos

Arpan-206:
**Checklist**:


rohitpaulk (changes requested):
Added note

PR #2941, opened 2025-07-07
Hide extension ideas if all extension ideas as released

Arpan-206:
**Checklist**:

### DIFF
--- a/app/controllers/roadmap/course-extension-ideas.ts
+++ b/app/controllers/roadmap/course-extension-ideas.ts
@@ -27,14 +27,9 @@ export default class CourseExtensionIdeasController extends Controller {
 
   get orderedCourses() {
     return this.model.courseExtensionIdeas
+      .rejectBy('developmentStatusIsReleased')
       .mapBy('course')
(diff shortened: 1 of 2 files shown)


app/controllers/roadmap/course-extension-ideas.ts
```diff
@@ -29,6 +29,12 @@ export default class CourseExtensionIdeasController extends Controller {
     return this.model.courseExtensionIdeas
       .mapBy('course')
       .uniq()
+      .filter((course) => {
+        // Check if this course has any extension ideas with status "not_started" or "in_progress"
+        const courseExtensionIdeas = this.model.courseExtensionIdeas.filterBy('course', course);
```
rohitpaulk, line 34: This is okay, but a simpler way to do this - add `rejectBy('releaseStatusIsReleased')` to the first line before `mapBy('course')`. In general prefer the shorthand methods unless they compromise readability, in general they're easier to quickly read through.

rohitpaulk (approved):

PR #2942, opened 2025-07-08
Use a drop down for challenge Selector

Arpan-206:
**Checklist**:


rohitpaulk (approved):
Comments added + styling tweak commits pushed

PR #2951, opened 2025-07-28
Fix dropdown colors

Arpan-206:
**Checklist**:

### DIFF
--- a/app/components/course-dropdown.hbs
+++ b/app/components/course-dropdown.hbs
@@ -23,7 +23,10 @@
       </button>
     </dd.Trigger>
     <dd.Content>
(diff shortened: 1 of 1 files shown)


app/components/course-dropdown.hbs
```diff
@@ -23,7 +23,10 @@
       </button>
     </dd.Trigger>
     <dd.Content>
-      <div class="bg-white dark:bg-gray-850 rounded-lg shadow-lg border dark:border-white/10 py-2 my-1" data-test-course-dropdown-content>
+      <div
+        class="bg-white dark:bg-gray-850 rounded-lg shadow-lg border border-gray-200 dark:border-white/10 py-2 mt-1 mb-1"
```
rohitpaulk, line 27:
```suggestion
        class="bg-white dark:bg-gray-850 rounded-lg shadow-lg border border-gray-200 dark:border-white/10 py-2 my-1"
```

(don't think we need to change that part?)

PR #2955, opened 2025-07-29
Pay page design overhaul

Arpan-206:
**Checklist**:

### DIFF
--- a/app/components/pay-page/choose-membership-plan-modal.hbs
+++ b/app/components/pay-page/choose-membership-plan-modal.hbs
@@ -27,7 +27,7 @@
         @isCreatingCheckoutSession={{this.isCreatingCheckoutSession}}
         @onBackButtonClick={{this.handleBackButtonClick}}
         @onExtraInvoiceDetailsToggle={{this.handleExtraInvoiceDetailsToggle}}
-        @onProceedToCheckout={{this.handleProceedToCheckoutClick}}
+        @onProceedToCheckout={{this.handleProceedToCheckoutButtonClick}}
         @regionalDiscount={{@regionalDiscount}}
         @selectedPlan={{this.selectedPlan}}
       />
--- a/app/components/pay-page/choose-membership-plan-modal.ts
+++ b/app/components/pay-page/choose-membership-plan-modal.ts
@@ -84,7 +84,7 @@ export default class ChooseMembershipPlanModal extends Component<Signature> {
   }
 
   @action
-  async handleProceedToCheckoutClick() {
+  async handleProceedToCheckoutButtonClick() {
     this.isCreatingCheckoutSession = true;
 
     const checkoutSession = this.store.createRecord('individual-checkout-session', {
@@ -93,7 +93,7 @@ export default class ChooseMembershipPlanModal extends Component<Signature> {
       pricingFrequency: this.selectedPlanId,
       promotionalDiscount: this.selectedPlanId === 'yearly' ? this.args.activeDiscountForYearlyPlan : null,
       regionalDiscount: this.args.regionalDiscount || null,
-      successUrl: `${window.location.origin}/tracks`,
+      successUrl: `${window.location.origin}/catalog`,
     });
 
     await checkoutSession.save();
--- a/app/components/pay-page/choose-membership-plan-modal/invoice-details-step.hbs
+++ b/app/components/pay-page/choose-membership-plan-modal/invoice-details-step.hbs
@@ -8,7 +8,7 @@
       @regionalDiscount={{@regionalDiscount}}
     />
 
-    <div class="absolute inset-0 bg-white/60 flex items-center justify-center opacity-100 transition-opacity duration-200">
+    <div class="absolute inset-0 bg-white/60 flex items-center justify-center">
       <TertiaryButton @size="small" {{on "click" @onBackButtonClick}}>
         Change plan
       </TertiaryButton>
--- a/app/components/pay-page/choose-membership-plan-modal/plan-card.hbs
+++ b/app/components/pay-page/choose-membership-plan-modal/plan-card.hbs
@@ -4,7 +4,7 @@
     rounded p-3
     {{if @isSelected 'bg-teal-50' 'bg-white hover:bg-gray-50'}}"
   role="button"
-  {{on "click" this.handleClick}}
+  {{on "click" @onClick}}
   data-test-plan-card
   ...attributes
 >
@@ -56,17 +56,13 @@
               <li data-test-promotional-discount-notice>
                 <b class="text-teal-500">{{@promotionalDiscount.percentageOff}}% off</b>
                 —
-                {{#if @promotionalDiscount.isFromAffiliateReferral}}
-                  <span class="text-nowrap">{{@promotionalDiscount.affiliateReferral.affiliateLink.usernameForDisplay}}'s</span>
-                  referral offer, expires in
-                  <span class="font-mono">{{this.promotionalDiscountCountdownTimerText}}</span>
-                {{else if @promotionalDiscount.isFromStage2Completion}}
-                  Stage 2 completion discount, expires in
-                  <span class="font-mono">{{this.promotionalDiscountCountdownTimerText}}</span>
-                {{else if @promotionalDiscount.isFromSignup}}
-                  Signup discount, expires in
-                  <span class="font-mono">{{this.promotionalDiscountCountdownTimerText}}</span>
-                {{/if}}
+                {{#if @promotionalDiscount.isFromAffiliateReferral}}<span
+                    class="text-nowrap"
+                  >{{@promotionalDiscount.affiliateReferral.affiliateLink.usernameForDisplay}}'s</span>
+                  referral offer{{else if @promotionalDiscount.isFromStage2Completion}}Stage 2 completion discount{{else if
+                  @promotionalDiscount.isFromSignup
+                }}Signup discount{{/if}}, expires in
+                <span class="font-mono">{{this.promotionalDiscountCountdownTimerText}}</span>
               </li>
             {{/if}}
           </ul>
--- a/app/components/pay-page/choose-membership-plan-modal/plan-card.ts
+++ b/app/components/pay-page/choose-membership-plan-modal/plan-card.ts
@@ -1,6 +1,5 @@
 import Component from '@glimmer/component';
 import { service } from '@ember/service';
-import { action } from '@ember/object';
 import type TimeService from 'codecrafters-frontend/services/time';
 import { formatTimeDurationForCountdown } from 'codecrafters-frontend/utils/time-formatting';
 import type RegionalDiscountModel from 'codecrafters-frontend/models/regional-discount';
@@ -55,11 +54,6 @@ export default class PlanCard extends Component<Signature> {
     }
 
     return formatTimeDurationForCountdown(this.args.promotionalDiscount.expiresAt, this.time.currentTime);
-  }
-
-  @action
-  handleClick() {
-    this.args.onClick();
   }
 }
 
--- a/app/components/pay-page/choose-membership-plan-modal/plan-selection-step.hbs
+++ b/app/components/pay-page/choose-membership-plan-modal/plan-selection-step.hbs
@@ -13,7 +13,7 @@
     <PrimaryButton @size="large" {{on "click" @onChoosePlanButtonClick}} data-test-choose-plan-button>
       Choose
       {{@selectedPlan.title}}
-      →
+      pass →
     </PrimaryButton>
   </div>
 
--- a/tests/pages/pay-page.js
+++ b/tests/pages/pay-page.js
@@ -12,9 +12,11 @@ export default create({
     clickOnProceedToCheckoutButton: clickable('[data-test-proceed-to-checkout-button]'),
     planCards: collection('[data-test-plan-card]', {
       discountedPriceText: text('[data-test-discounted-price]'),
+
       promotionalDiscountNotice: {
         scope: '[data-test-promotional-discount-notice]',
       },
+
       regionalDiscountNotice: {
         scope: '[data-test-regional-discount-notice]',
       },
--- a/app/components/pay-page/choose-membership-plan-modal/extra-invoice-details-toggle.hbs
+++ b/app/components/pay-page/choose-membership-plan-modal/extra-invoice-details-toggle.hbs
@@ -16,9 +16,9 @@
     data-test-extra-invoice-details-toggle
   >
     <span class="flex grow flex-col">
-      <span class="text-sm font-semibold leading-6 {{if @extraInvoiceDetailsRequested 'text-teal-600' 'text-gray-900'}}" id="availability-label">I
-        need extra details on my invoice</span>
-      <span class="text-xs text-gray-400" id="availability-description">
+      <span class="text-sm font-semibold leading-6 {{if @extraInvoiceDetailsRequested 'text-teal-600' 'text-gray-900'}}">I need extra details on my
+        invoice</span>
+      <span class="text-xs text-gray-400">
         Choose this option if you need extra details like tax ID, business address etc. on your invoice.
       </span>
     </span>


Arpan-206:
@rohitpaulk thought you might wanna review now that's hooked up to real data (still gonna wait for percy + a lot more of my own manual test before merge tho)

rohitpaulk:
Just testing functionality/design for now, will review code later once your testing is done + you confirm everything is ready for review (don't need to wait for percy): 

- Clicking on "continue with free ->" triggers login for me, it shouldn't - should just navigate to the catalog without auth
  - Same with view plans, a user must be able to view plans without logging in just like they can today. We don't want to restrict viewing prices
  - Same with teams/pay, user must be able to go to that page without logging in
- Flow when authed seems okay! I wasn't able to test with discounts and stuff but the regular flow worked fine

rohitpaulk (changes requested):
Functionality notes added!

app/components/pay-page/choose-membership-plan-modal.ts
```diff
@@ -0,0 +1,134 @@
...
+
+export default class ChooseMembershipPlanModal extends Component<Signature> {
+  @service declare authenticator: AuthenticatorService;
+  @service declare store: Store;
+
+  transition = fade;
+
+  @tracked extraInvoiceDetailsRequested = false;
+  @tracked isCreatingCheckoutSession = false;
+  @tracked previewType: 'plan' | 'invoice-details' = 'plan';
```
rohitpaulk, line 31:
```suggestion
  @tracked previewType: 'plan' | 'invoice-details' = 'plan';
```

@Arpan-206 I don't put a lot of thought into variable names etc. in design previews, upto you to clean them up. `previewType` doesn't really make sense here - check other wizard-like flows and I think we use `currentStep` in most of them. The `teams/pay` one is one example of a wizard flow, the onboarding survey is another.

app/components/pay-page/choose-membership-plan-modal.hbs
```diff
@@ -0,0 +1,173 @@
...
+    </div>
+
+    <p class="text-gray-700 text-sm text-center text-balance mt-2">
+      CodeCrafters members routinely unlock promotions and 5-figure salary jumps.
+    </p>
+  </div>
+
+  <AnimatedContainer>
+    {{#animated-if (eq this.previewType "plan") use=this.transition duration=200}}
+      <div>
```
rohitpaulk, line 14: Extremely long file, let's split this into components

app/components/pay-page/choose-membership-plan-modal.ts
```diff
@@ -0,0 +1,134 @@
...
+import type PromotionalDiscountModel from 'codecrafters-frontend/models/promotional-discount';
+import type AuthenticatorService from 'codecrafters-frontend/services/authenticator';
+
+interface Signature {
+  Element: HTMLAnchorElement;
+
+  Args: {
+    onClose: () => void;
+    regionalDiscount?: RegionalDiscountModel | null;
+    shouldApplyRegionalDiscount?: boolean;
```
rohitpaulk, line 18: Let's remove this and always apply a regional discount - this design isn't intended to be shipped with "optional" regional discounts, it's very easy to miss the initial banner and users will complain later on

app/components/pay-page/choose-membership-plan-modal.ts
```diff
@@ -0,0 +1,134 @@
...
+  get selectedPlanText() {
+    switch (this.selectedPlan) {
+      case '3-month':
+        return '3 month pass';
+      case '1-year':
+        return '1 year pass';
+      case 'lifetime':
+        return 'Lifetime pass';
+      default:
+        return '3 month pass';
```
rohitpaulk, line 87: Raise an error / notify sentry - this shouldn't be happening

app/components/pay-page/choose-membership-plan-modal.ts
```diff
@@ -0,0 +1,134 @@
...
+      regionalDiscount: this.args.shouldApplyRegionalDiscount ? this.args.regionalDiscount || null : null,
+      successUrl: `${window.location.origin}/settings/billing`,
+    });
+
+    await checkoutSession.save();
+    window.location.href = checkoutSession.url;
+  }
+
+  @action
+  selectPlan(plan: '3-month' | '1-year' | 'lifetime') {
```
rohitpaulk, line 125: Consistency: `handlePlanSelect` / `handlePlanSelection`

rohitpaulk (changes requested):
Added one round of comments - @Arpan-206 definitely needs more thought on structure, naming etc. Don't rely heavily on design preview PRs for anything except for design

app/components/pay-page/choose-membership-plan-modal.ts
```diff
@@ -0,0 +1,108 @@
...
+  { id: 'quarterly', title: '3 month pass', priceInDollars: 120, validityInMonths: 3 },
+  { id: 'yearly', title: '1 year pass', priceInDollars: 360, validityInMonths: 12 },
+  { id: 'lifetime', title: 'Lifetime pass', priceInDollars: 1490 },
+];
+
+interface Signature {
+  Element: HTMLDivElement;
+
+  Args: {
+    activeDiscountForYearlyPlan?: PromotionalDiscountModel | null;
```
rohitpaulk, line 30: Let's use either `?` or `| null`, not both - the ultimate signature here would be `Model | null | undefined` if we did this, better to restrict to one so it's easier to deal with within the component.

app/components/pay-page/choose-membership-plan-modal.ts
```diff
@@ -0,0 +1,108 @@
...
+}
+
+export default class ChooseMembershipPlanModal extends Component<Signature> {
+  @service declare authenticator: AuthenticatorService;
+  @service declare store: Store;
+
+  transition = fade;
+  pricingPlans = PRICING_PLANS;
+
+  @tracked extraInvoiceDetailsRequested = false;
```
rohitpaulk, line 43: Sort properties alphabetically

app/components/pay-page/choose-membership-plan-modal.ts
```diff
@@ -0,0 +1,134 @@
...
+    this.previewType = 'invoice-details';
+  }
+
+  @action
+  async handleProceedToCheckoutButtonClick(pricingFrequency: 'quarterly' | 'yearly' | 'lifetime') {
+    if (!this.authenticator.isAuthenticated) {
+      this.authenticator.initiateLoginAndRedirectTo(`${window.location.origin}/pay`);
+
+      return;
+    }
```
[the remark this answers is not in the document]
rohitpaulk: We're okay with this

app/components/pay-page/choose-membership-plan-modal.ts
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+  @action
+  handlePlanSelect(plan: PricingPlan) {
+    this.selectedPlanId = plan.id;
+  }
+
+  @action
+  async handleProceedToCheckoutClick() {
+    if (!this.authenticator.isAuthenticated) {
```
rohitpaulk, line 82: Let's do this on when a person chooses a plan instead of when they to the "invoice details" thing (that way they only lose one step worth of progress when they hit the page again, doesnt' seem like they filled in a bunch of stuff and then lost context)

app/components/pay-page/choose-membership-plan-modal/plan-card.hbs
```diff
@@ -0,0 +1,110 @@
...
+          <span class="inline-block bg-teal-500 text-white font-bold rounded px-1.5 py-1 text-xxs uppercase">best value</span>
+        {{/if}}
+      </div>
+
+      <div class="prose prose-xs prose-compact mt-1 text-gray-500">
+        {{#if @plan.validityInMonths}}
+          <p>
+            Access to all membership benefits for
+            {{@plan.validityInMonths}}
+            {{if (eq @plan.validityInMonths 1) "month" "months"}}
```
rohitpaulk, line 35: For the yearly plan we want this to say 1 year, not 12 months

app/components/pay-page/choose-membership-plan-modal/plan-card.ts
```diff
@@ -0,0 +1,73 @@
...
+    let discountedPrice = basePrice;
+
+    // Apply promotional discount first (yearly plans only)
+    if (this.args.activeDiscountForYearlyPlan && this.args.plan.id === 'yearly') {
+      discountedPrice = this.args.activeDiscountForYearlyPlan.computeDiscountedPrice(basePrice);
+    }
+
+    // Apply regional discount
+    if (this.args.regionalDiscount) {
+      const regionalDiscountAmount = (discountedPrice * this.args.regionalDiscount.percentOff) / 100;
```
rohitpaulk, line 44: Let's add `computeDiscountedPrice` to regionalDiscount too, to be symmetric with `PromotionalDiscountModel`, that'll make this similar to the above line to read

app/components/pay-page/choose-membership-plan-modal/plan-card.hbs
```diff
@@ -0,0 +1,110 @@
...
+                <b class="text-teal-500">{{@regionalDiscount.percentOff}}% off</b>
+                —
+                {{@regionalDiscount.countryName}}
+                discount
+              </li>
+            {{/if}}
+            {{#if (eq @plan.id "yearly")}}
+              {{#if
+                (and
+                  @activeDiscountForYearlyPlan @activeDiscountForYearlyPlan.isFromAffiliateReferral (not @activeDiscountForYearlyPlan.isFromSignup)
```
rohitpaulk, line 58:
```suggestion
                  @activeDiscountForYearlyPlan @activeDiscountForYearlyPlan.isFromAffiliateReferral
```

If `@activeDiscountForYearlyPlan` is from an affiliate referral, it naturally won't be from a signup.

app/components/pay-page/choose-membership-plan-modal/plan-card.hbs
```diff
@@ -0,0 +1,110 @@
...
+          <ul class="mt-0">
+            {{#if @regionalDiscount}}
+              <li data-test-modal-regional-discount-notice>
+                <b class="text-teal-500">{{@regionalDiscount.percentOff}}% off</b>
+                —
+                {{@regionalDiscount.countryName}}
+                discount
+              </li>
+            {{/if}}
+            {{#if (eq @plan.id "yearly")}}
```
rohitpaulk, line 55: There's a bit of duplication of logic here, we're maintaining the fact that certain discounts only apply to the yearly plan both here and in the place where we calculate the discounted price. 

Idea to clean this up: 

- Add `get applicableDiscounts()`, which just returns [](PromotionalDiscount | RegionalDiscount)
- Use that in the `discountedPriceInDollars` calculation, and also here use it as `{{#each this.applicableDiscounts}}`.

This ensures that whatever is used in the calculation will be handled here too, without needing to add any extra logic in the template

app/components/pay-page/choose-membership-plan-modal/plan-card.ts
```diff
@@ -0,0 +1,73 @@
...
+import type PromotionalDiscountModel from 'codecrafters-frontend/models/promotional-discount';
+import type { PricingPlan } from '../choose-membership-plan-modal';
+
+interface Signature {
+  Element: HTMLDivElement;
+
+  Args: {
+    activeDiscountForYearlyPlan?: PromotionalDiscountModel | null;
+    isSelected: boolean;
+    onPlanSelect: (plan: PricingPlan) => void;
```
rohitpaulk, line 16:
```suggestion
    onClick: () => void;
```

At the plan-card level, this action is just "onClick". A user can't select a different plan from this card so we can keep the action more simple

tests/pages/pay-page.js
```diff
@@ -1,16 +1,34 @@
...
+
+  clickOnChoosePlanButton: clickable('[data-test-choose-plan-button]'),
   clickOnExtraInvoiceDetailsToggle: clickable('[data-test-extra-invoice-details-toggle]'),
   clickOnProceedToCheckoutButton: clickable('[data-test-proceed-to-checkout-button]'),
-  clickOnStartPaymentButtonForMonthlyPlan: clickable('[data-test-monthly-pricing-card] [data-test-start-payment-button]'),
-  clickOnStartPaymentButtonForYearlyPlan: clickable('[data-test-yearly-pricing-card] [data-test-start-payment-button]'),
+
   header: Header,
 
+  pricingPlanCards: collection('[data-test-pricing-plan-card]', {
```
rohitpaulk, line 14: These properties are sorted alphabetically

tests/pages/pay-page.js
```diff
@@ -1,16 +1,34 @@
...
+    ctaButton: {
+      scope: '[data-test-pricing-plan-card-cta]',
+    },
+  }),
+
+  chooseMembershipPlanModal: {
+    planCards: collection('[data-test-modal-plan-card]', {
```
rohitpaulk, line 21:
```suggestion
    planCards: collection('[data-test-plan-card]', {
```

Add `scope: 'data-test-choose-membership-plan-modal'` to this and we won't have to keep qualifying each `data-test` property as `data-test-modal` (it'll automatically use nested to only find out relevant items

tests/pages/pay-page.js
```diff
@@ -1,16 +1,34 @@
...
+      discountedPriceText: text('[data-test-discounted-price]'),
+      discountNotice: {
+        scope: '[data-test-discount-notice]',
```
rohitpaulk, line 24: If we've got `data-test-discount-notice` and `data-test-regional-discount-notice`, it isn't very clear how these are different. Either we should do `data-test-promotional-discount-notice` and `data-test-regional-discount-notice`, or we should just do `discountNotices: collection('[data-test-discount-notice]')` and assert on the text of the notices instead

tests/pages/pay-page.js
```diff
@@ -1,16 +1,34 @@
-import { clickable, collection, create, text, triggerable, visitable } from 'ember-cli-page-object';
+import { clickable, collection, create, text, visitable } from 'ember-cli-page-object';
 import AccountDropdown from 'codecrafters-frontend/tests/pages/components/account-dropdown';
 import Header from 'codecrafters-frontend/tests/pages/components/header';
 
 export default create({
   accountDropdown: AccountDropdown,
-  clickOnApplyRegionalDiscountButton: clickable('[data-test-apply-regional-discount-button]'),
+
```
rohitpaulk, line 7: Let's stick to convention and not add spaces unless there's a nested block that needs spaces around it (all single-line properties don't need spaces around them)

tests/pages/pay-page.js
```diff
@@ -1,16 +1,34 @@
...
+  clickOnChoosePlanButton: clickable('[data-test-choose-plan-button]'),
```
rohitpaulk, line 8: All of this (choose plan, extra invoice details, proceed to checkout etc.) should be nested inside chooseMembershipPlanModal, not placed directly on the pay-page object

app/components/pay-page/choose-membership-plan-modal/plan-card.ts
```diff
@@ -0,0 +1,73 @@
...
+import { formatTimeDurationForCountdown } from 'codecrafters-frontend/utils/time-formatting';
+import type RegionalDiscountModel from 'codecrafters-frontend/models/regional-discount';
+import type PromotionalDiscountModel from 'codecrafters-frontend/models/promotional-discount';
+import type { PricingPlan } from '../choose-membership-plan-modal';
+
+interface Signature {
+  Element: HTMLDivElement;
+
+  Args: {
+    activeDiscountForYearlyPlan?: PromotionalDiscountModel | null;
```
rohitpaulk, line 14: If we already know what plan we're passing in, we can just pass in `promotionalDiscount` and `regionalDiscount` (just pass in `promotionalDiscount` as empty from the caller). 

We don't need to pass in `activeDiscountForYearlyPlan` and then filter out applicability here since all plan cards other than the yearly plan card don't need access to the active discount for the yearly plan

app/components/pay-page/choose-membership-plan-modal.hbs
rohitpaulk: @Arpan-206 think this was missed, I'd recommend splitting up the steps here into different components

app/components/pay-page/choose-membership-plan-modal/invoice-details-step.hbs
```diff
@@ -0,0 +1,45 @@
...
+  <div class="relative">
+    <PayPage::ChooseMembershipPlanModal::PlanCard
+      @isSelected={{true}}
+      @onClick={{(noop)}}
+      @plan={{@selectedPlan}}
+      @promotionalDiscount={{if (eq @selectedPlan.id "yearly") @activeDiscountForYearlyPlan null}}
+      @regionalDiscount={{@regionalDiscount}}
+    />
+
+    <div class="absolute inset-0 bg-white/60 flex items-center justify-center opacity-100 transition-opacity duration-200">
```
rohitpaulk, line 11:
```suggestion
    <div class="absolute inset-0 bg-white/60 flex items-center justify-center">
```

The design preview initially used animations, but we aren't using those anymore and instead use a 60% opacity background (`bg-white/60`)

app/components/pay-page/choose-membership-plan-modal/plan-card.hbs
```diff
@@ -0,0 +1,100 @@
...
+                —
+                {{#if @promotionalDiscount.isFromAffiliateReferral}}
+                  <span class="text-nowrap">{{@promotionalDiscount.affiliateReferral.affiliateLink.usernameForDisplay}}'s</span>
+                  referral offer, expires in
+                  <span class="font-mono">{{this.promotionalDiscountCountdownTimerText}}</span>
+                {{else if @promotionalDiscount.isFromStage2Completion}}
+                  Stage 2 completion discount, expires in
+                  <span class="font-mono">{{this.promotionalDiscountCountdownTimerText}}</span>
+                {{else if @promotionalDiscount.isFromSignup}}
+                  Signup discount, expires in
```
rohitpaulk, line 67: Could move the whole expiry thing out here and place it below: 

```ts
, expires in <font-mono>....</blah>
```

This is similar to how the first portion above is constant too (`percentOff% -`)

app/components/pay-page/choose-membership-plan-modal/plan-card.ts
```diff
@@ -0,0 +1,70 @@
...
+    if (!this.args.promotionalDiscount) {
+      return '';
+    }
+
+    return formatTimeDurationForCountdown(this.args.promotionalDiscount.expiresAt, this.time.currentTime);
+  }
+
+  @action
+  handleClick() {
+    this.args.onClick();
```
rohitpaulk, line 62: For simple cases like this you can just do `{{on "click" @onClick}}` in the template. Only use handlers when there's a reason to add some logic

app/components/pay-page/choose-membership-plan-modal/plan-selection-step.hbs
```diff
@@ -0,0 +1,29 @@
...
+      <PayPage::ChooseMembershipPlanModal::PlanCard
+        @isSelected={{eq @selectedPlan.id plan.id}}
+        @onClick={{fn @onPlanSelect plan}}
+        @plan={{plan}}
+        @promotionalDiscount={{if (eq plan.id "yearly") @activeDiscountForYearlyPlan null}}
+        @regionalDiscount={{@regionalDiscount}}
+      />
+    {{/each}}
+
+    <PrimaryButton @size="large" {{on "click" @onChoosePlanButtonClick}} data-test-choose-plan-button>
```
rohitpaulk, line 13: This should say "Choose 3 months pass ->" btw, not just "Choose 3 months ->"

tests/pages/pay-page.js
```diff
@@ -1,16 +1,38 @@
...
-  clickOnApplyRegionalDiscountButton: clickable('[data-test-apply-regional-discount-button]'),
-  clickOnExtraInvoiceDetailsToggle: clickable('[data-test-extra-invoice-details-toggle]'),
-  clickOnProceedToCheckoutButton: clickable('[data-test-proceed-to-checkout-button]'),
-  clickOnStartPaymentButtonForMonthlyPlan: clickable('[data-test-monthly-pricing-card] [data-test-start-payment-button]'),
-  clickOnStartPaymentButtonForYearlyPlan: clickable('[data-test-yearly-pricing-card] [data-test-start-payment-button]'),
+
+  chooseMembershipPlanModal: {
+    scope: '[data-test-choose-membership-plan-modal]',
+    clickOnChoosePlanButton: clickable('[data-test-choose-plan-button]'),
+    clickOnExtraInvoiceDetailsToggle: clickable('[data-test-extra-invoice-details-toggle]'),
```
rohitpaulk, line 11: Same note as previously added - Let's stick to convention and not add spaces unless there's a nested block that needs spaces around it (all single-line properties don't need spaces around them)

rohitpaulk (approved):

PR #2957, opened 2025-07-30
Using github.com to view full code for any code example (Frontend PR)

Arpan-206:
**Checklist**:


tests/acceptance/course-page/code-examples/export-to-github-test.js
```diff
@@ -0,0 +1,173 @@
...
+import { setupApplicationTest } from 'codecrafters-frontend/tests/helpers';
+import { signIn } from 'codecrafters-frontend/tests/support/authentication-helpers';
+import { settled } from '@ember/test-helpers';
+import courseOverviewPage from 'codecrafters-frontend/tests/pages/course-overview-page';
+
+module('Acceptance | course-page | code-examples | export-to-github', function (hooks) {
+  setupApplicationTest(hooks);
+  setupAnimationTest(hooks);
+
+  async function setupTestScenario(server) {
```
rohitpaulk, line 18: The pattern to do this in q-unit is to do `hooks.beforeEach` (search the codebase to find usages)

tests/acceptance/course-page/code-examples/export-to-github-test.js
```diff
@@ -0,0 +1,173 @@
...
+    const redis = server.schema.courses.findBy({ slug: 'redis' });
+    const python = server.schema.languages.findBy({ slug: 'python' });
+
+    server.create('repository', 'withFirstStageCompleted', {
+      course: redis,
+      language: python,
+      user: currentUser,
+    });
+
+    return {
```
rohitpaulk, line 38: This returns 5 different values, and only one of them actually gets used - could just have been `return createCommunityCourse....`

tests/acceptance/course-page/code-examples/export-to-github-test.js
```diff
@@ -0,0 +1,173 @@
...
+    await coursePage.sidebar.clickOnStepListItem('Respond to PING');
+    await coursePage.clickOnHeaderTabLink('Code Examples');
+  }
+
+  test('creates new export when none exists', async function (assert) {
+    const { solution } = await setupTestScenario.call(this, this.server);
+
+    let exportCreateCalled = false;
+    this.server.post('/community-solution-exports', function (schema) {
+      exportCreateCalled = true;
```
rohitpaulk, line 60: @Arpan-206 I'd like to understand your thought process here a bit more - can you find me 3 other places in the codebase where we test creating a resource, and how they handle testing that action?

tests/acceptance/course-page/code-examples/export-to-github-test.js
```diff
@@ -0,0 +1,173 @@
...
+        githubRepositoryUrl: 'https://github.com/test-user/test-repo',
+        communitySolution: solution,
+      });
+    });
+
+    await navigateToCodeExamples();
+    await percySnapshot('Code Examples - View on GitHub button (unpublished solution)');
+
+    await codeExamplesPage.solutionCards[0].highlightedFileCards[0].clickOnViewOnGithubButton();
+    await settled();
```
rohitpaulk, line 74: Let's test the actual user-facing action too, i.e. the redirect

tests/acceptance/course-page/code-examples/export-to-github-test.js
```diff
@@ -0,0 +1,173 @@
...
+    this.server.post('/community-solution-exports', function () {
+      return new Response(500, {}, { error: 'Export creation failed' });
+    });
+
+    await navigateToCodeExamples();
+
+    await codeExamplesPage.solutionCards[0].highlightedFileCards[0].clickOnViewOnGithubButton();
+    await settled();
+
+    assert.dom('[data-test-view-on-github-button]').exists('view button should still exist after failure');
```
rohitpaulk, line 171: Use page objects, not `assert.dom`. We use `assert.dom` in very few places and only when the use-case can't be achieved via page objects.

rohitpaulk: This is another example where just searching the codebase will give you an idea of what patterns we follow. Look for `assert.dom` usages and you'll find very few

app/components/course-page/course-stage-step/community-solution-card/github-file-actions.ts
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    this.isCreatingExport = true;
+
+    try {
+      const exportRecord = await this.args.solution.createExport();
+      this.isCreatingExport = false;
+      const githubUrl = exportRecord.githubUrlForFile(this.args.filename);
+
+      if (exportRecord.status === 'provisioned') {
```
rohitpaulk, line 49: If a created export's status isn't provisioned, we should be raising an error - it's an invalid condition. I'd remove this condition entirely - it's a contract the backend needs to satisfy

app/components/course-page/course-stage-step/community-solution-card/github-file-actions.ts
```diff
@@ -0,0 +1,80 @@
...
+  @action
+  handleViewOnGithubButtonClick() {
+    const latestExport = this.getLatestUnexpiredExport();
+
+    if (latestExport?.status === 'provisioned') {
+      const githubUrl = latestExport.githubUrlForFile(this.args.filename);
+      latestExport.markAsAccessed({});
+      window.open(githubUrl, '_blank', 'noopener,noreferrer');
+    } else {
+      this.createExport();
```
rohitpaulk, line 71: The `window.open` code would fit better here than inside `createExport`, reasons: 

- It's parallel to the `window.open` above so makes it clear that in both cases we're redirecting the user somewhere
- Makes the `createExport` name a bit more appropriate - today it handles create and redirect, and once the redirect is moved here it'll handle only the create part

app/components/course-page/course-stage-step/community-solution-card/github-file-actions.ts
```diff
@@ -0,0 +1,80 @@
...
+        window.open(githubUrl, '_blank', 'noopener,noreferrer');
+      }
+    } catch (error) {
+      Sentry.captureException(error);
+      this.isCreatingExport = false;
+    }
+  }
+
+  private getLatestUnexpiredExport(): CommunitySolutionExportModel | null {
+    return this.args.solution.exports?.reject((exportRecord) => new Date() >= exportRecord.expiresAt)?.sortBy('expiresAt')?.lastObject || null;
```
rohitpaulk, line 59:
```suggestion
    return this.args.solution.exports.rejectBy('isExpired').sortBy('expiresAt').lastObject || null;
```

(None of these values can be null in the middle)

app/components/course-page/course-stage-step/community-solution-card/github-file-actions.ts
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  @action
+  handleViewOnGithubButtonClick() {
+    const latestExport = this.getLatestUnexpiredExport();
```
rohitpaulk, line 64:
```suggestion
    const latestExport = this.getLatestProvisionedExport();
```

One way to simplify this code would be to move the `isProvisioned` check into `getLatestProvisionedExport` - that way all the filtering stays there, and the if condition here is just `if (latestExport)`

rohitpaulk (changes requested):
Added another round of comments

Arpan-206:
@rohitpaulk adding another thing, just saw the frontend modal and this many email suffixes will look badddd on the UI 
Screenshot 2025-08-13 at 17 25 26

So we need to either add wildcard support or support all subdomains of a domain

rohitpaulk (approved):
(We'll look over the test file tomorrow)

PR #3006, opened 2025-08-14
Initiate Login on Campus Claim Offer Button for Non Logged in chaps

Arpan-206:
**Checklist**:


app/components/institution-page/claim-offer-button.ts
```diff
@@ -29,6 +29,11 @@ export default class ClaimOfferButton extends Component<Signature> {
 
   @action
   handleClick() {
+    if (!this.authenticator.isAuthenticated) {
```
rohitpaulk, line 32: @Arpan-206 tests - this will be a very common flow

PR #3007, opened 2025-08-14
Add Pre-Rendering to Institutions

Arpan-206:
**Checklist**:


vercel-functions/prerender/campus/[institution_slug].func/.vc-config.json
```diff
@@ -0,0 +1,9 @@
+{
```
Comment (rohitpaulk) [inline vercel-functions/prerender/campus/[institution_slug].func/.vc-config.json:1]: @Arpan-206 this whole folder needs to be a symlink - example: 

Screenshot 2025-08-14 at 20 22 51


Only the `prerender-config.json` file needs to be maintained separately.

rohitpaulk (changes requested):
Note added

rohitpaulk (approved):

PR #3043, opened 2025-08-21
Add CampusBadge component and integrate it into billing status display logic.

Arpan-206:
**Checklist**:

### DIFF
--- a/tests/pages/components/billing-status-badge.js
+++ b/tests/pages/components/billing-status-badge.js
@@ -1,6 +1,11 @@
 import { hasClass, text, triggerable } from 'ember-cli-page-object';
 
 export default {
+  campusBadge: {
+    hover: triggerable('mouseenter'),
+    scope: '[data-test-campus-badge]',
+  },
+
   discountTimerBadge: {
     scope: '[data-test-discount-timer-badge]',
     hover: triggerable('mouseenter'),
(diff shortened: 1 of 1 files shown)


app/models/user.ts
```diff
@@ -102,7 +104,13 @@ export default class UserModel extends Model {
   }
 
   get canAccessMembershipBenefits() {
-    return this.isVip || this.hasActiveSubscription || this.teamHasActiveSubscription || this.teamHasActivePilot;
+    return (
+      this.isVip ||
+      this.hasActiveInstitutionMembershipGrants ||
```
rohitpaulk, line 109: We've tied membership grants to subscriptions for a reason - so that we don't need to add special cases like this. Every membership grant has an associated subscription. 

Shouldn't have to edit this at all

mirage/config.js
```diff
@@ -105,6 +105,10 @@ export default function (config) {
           parentComment: belongsTo('community-course-stage-solution-comment', { inverse: null }),
         }),
         fakeLogstream: Model.extend({}),
+        institutionMembershipGrant: Model.extend({
```
rohitpaulk, line 108: Are you sure we need this? Can't these relationships not be inferred? The other ones in this file use relationships that cannot be inferred for some reason (I forget exactly why but I think currentUserDownvotes / currentUserUpvotes is the reason)

tests/pages/components/billing-status-badge.js
```diff
@@ -14,6 +14,11 @@ export default {
     scope: '[data-test-free-weeks-left-button]',
   },
 
+  campusBadge: {
```
rohitpaulk, line 17: Alphabetical ordering

rohitpaulk (approved):

PR #3050, opened 2025-08-23
Pre fill email addresses from GitHub profile

Arpan-206:
**Checklist**:

### DIFF
--- a/app/components/institution-page/campus-program-application-modal.ts
+++ b/app/components/institution-page/campus-program-application-modal.ts
@@ -28,7 +28,7 @@ export default class CampusProgramApplicationModal extends Component<Signature> 
 
   get firstMatchingInstitutionalEmailAddress(): EmailAddressModel | null {
     return (
-      this.store.peekAll('email-address').find((emailAddress) => {
+      this.authenticator.currentUser!.emailAddresses.find((emailAddress) => {
         return this.args.institution.officialEmailAddressSuffixes.some((suffix) => emailAddress.value.endsWith(suffix));
       }) || null
     );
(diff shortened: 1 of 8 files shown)


app/models/email-address.ts
```diff
@@ -0,0 +1,5 @@
+import Model, { attr } from '@ember-data/model';
+
+export default class EmailAddressModel extends Model {
+  @attr('string') declare emailAddress: string;
```
rohitpaulk, line 4:
```suggestion
  @attr('string') declare value: string;
```

(Same note as in core)

app/components/institution-page/campus-program-application-modal.ts
```diff
@@ -18,11 +20,20 @@ interface Signature {
...
   @service declare authenticator: AuthenticatorService;
+  @service declare store: Store;
 
   @tracked isOverridingSavedGrantApplication = false;
 
   transition = fade;
 
+  get firstMatchingInstitutionalEmailAddress(): EmailAddressModel | null {
+    return (
+      this.store.peekAll('email-address').find((emailAddress) => {
```
rohitpaulk, line 31: Let's do `authenticator.currentUser.emailAddresses` instead of using `store.peekAll` - whenever dealing with resources that belong to users and aren't system-wide, we should be very careful about using `peekAll`. This can unintentionally grant a user access to resources that belong to someone else if they're loaded in the store. Even though today we only return email addresses for the current user, this might change in the future and it's not going to be obvious that this call needs to change. 

You'll see that almost all `peekAll('')` calls are for resources that are system-wide and not user-specific, like courses, languages, concepts. etc: 

Screenshot 2025-08-24 at 00 41 21

rohitpaulk (approved):

PR #3120, opened 2025-09-21
Show completed stage indicators on Course Overview page

### LINKED ISSUE #3095: Course overview page should show ticked stages just like track page
When viewing a challenge I've already started, the overview page looks the same except for the "Resume" button: 

Image

Let's instead change this to match what the tracks page does: 

Image

VasylMarchuk:
Closes #3095

### Brief

This adds complete stage indicators to the Course Overview page

### Screenshot

Знімок екрана 2025-09-21 о 17 57 16

### Checklist

---

> [!NOTE]
> Displays completion checkmarks for stages on the Course Overview page by computing completed stages in the route and passing them through component hierarchy.
> 
> - **Course Overview**
>   - Compute `completedStages` in `routes/course-overview.ts` via `#peekCompletedStages()` (from `course-stage-completion` records) and include in model.
>   - Pass `@completedStages` through `templates/course-overview.hbs` → `CourseOverviewPage::IntroductionAndStages` → `CourseOverviewPage::StageList` → `TrackPage::CourseCard::StageList`.
>   - Make `completedStages` args optional in `StageList` components (`course-overview-page` and `track-page/course-card`).
> - **UI**
>   - Add `data-test-stage-complete-checkmark` wrapper around the checkmark in `stage-list-item.hbs`.
> - **Tests**
>   - Update acceptance test to assert checkmarks/difficulty labels on overview and add Percy snapshot.
>   - Extend `tests/pages/course-overview-page.js` with `isPresent` checks for completion and difficulty labels.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 13d0e0ede1c58ecce74f3f1120e43031664fca57. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/app/controllers/course-overview.ts
+++ b/app/controllers/course-overview.ts
@@ -29,7 +29,7 @@ export default class CourseOverviewController extends Controller {
       return [];
     }
 
-    return this.userRepositories.filter((repository) => repository.course === this.model.course).flatMap((repository) => repository.completedStages);
+    return this.userRepositories.flatMap((repository) => repository.completedStages);
   }
 
   get currentUser() {
--- a/app/controllers/course-overview.ts
+++ b/app/controllers/course-overview.ts
@@ -25,10 +25,6 @@ export default class CourseOverviewController extends Controller {
   }
 
   get completedStages() {
-    if (!this.model.course) {
-      return [];
-    }
-
     return this.userRepositories.flatMap((repository) => repository.completedStages);
   }
 
--- a/app/controllers/course-overview.ts
+++ b/app/controllers/course-overview.ts
@@ -24,6 +24,14 @@ export default class CourseOverviewController extends Controller {
     }
   }
 
+  get completedStages() {
+    if (!this.model.course) {
+      return [];
+    }
+
+    return this.userRepositories.filter((repository) => repository.course === this.model.course).flatMap((repository) => repository.completedStages);
+  }
+
   get currentUser() {
     return this.authenticator.currentUser;
   }
--- a/app/routes/course-overview.ts
+++ b/app/routes/course-overview.ts
@@ -9,9 +9,11 @@ import type MetaDataService from 'codecrafters-frontend/services/meta-data';
 import type Store from '@ember-data/store';
 import type RouterService from '@ember/routing/router-service';
 import { inject as service } from '@ember/service';
+import type CourseStageModel from 'codecrafters-frontend/models/course-stage';
 
 export interface ModelType {
   course: CourseModel;
+  completedStages?: CourseStageModel[];
 }
 
 export default class CourseOverviewRoute extends BaseRoute {
@@ -52,8 +54,11 @@ export default class CourseOverviewRoute extends BaseRoute {
         include: 'extensions,stages,language-configurations.language',
       });
 
+      const course = this.store.peekAll('course').findBy('slug', params.course_slug);
+
       return {
-        course: this.store.peekAll('course').findBy('slug', params.course_slug),
+        course,
+        completedStages: this.#peekCompletedStages(course),
       };
     } else {
       const courses = await this.store.findAll('course', {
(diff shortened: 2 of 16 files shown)


VasylMarchuk:
Percy prevents tests from succeeding, [here's the build with a new screenshot](https://percy.io/CodeCrafters/web/frontend/builds/43443698/changed/2308033231?aiToggleEnabled=enabled&browser=chrome&browser_ids=69&group_snapshots_by=similar_diff&subcategories=approved&test_case_ids=&viewLayout=overlay&viewMode=new&width=1280&widths=375%2C1280)

app/routes/course-overview.ts
```diff
@@ -73,7 +78,16 @@ export default class CourseOverviewRoute extends BaseRoute {
...
 
-      return { course };
+      return {
+        course,
+        completedStages: this.#peekCompletedStages(course),
+      };
     }
   }
+
+  #peekCompletedStages(course: CourseModel) {
```
rohitpaulk, line 88: Can this be directly picked off `authenticator.currentUser` instead of peeking globally? 

i.e. something like: 

```javascript
this.authenticator.currentUser
  .repositories
  .filter((repository) => repository.course === course)
  .flatMap('courseStageCompletions')
  .mapBy('courseStage')
```

In general we try to avoid `peekAll` usages unless it's a system-wide resource that doesn't belong to a user (like `course`, `concept` etc.). With most user-related resources there's usually a relationship that can get to the resource we want. This also prevents bugs in the future if we pull in course stage completions from other users (it'll end up showing stages as completed when they aren't).

rohitpaulk (approved):

app/routes/course-overview.ts
VasylMarchuk: I initially tried adding a `@hasMany('courseStageCompletions')` relationship to the `CourseStage` model, and properly specifying the inverse for `@belongsTo('courseStage')` in `CourseStageCompletion` model. This way I could access completions simply via `course.stages.completions`.

But the backend isn't sending `course_stage_completions` for `CourseStage`s and I couldn't figure out the proper `include` string:

```ts
this.store.findAll('course', {
  include: 'extensions,stages,language-configurations.language',
})
```

I'll try use the approach you suggested, but maybe doing it via models/relationships is a better way, @rohitpaulk ?

VasylMarchuk: Changed to the suggested method and it's working. Also refactored the model hook a little bit to simplify and de-duplicate the logic.

VasylMarchuk: In addition, moved `completedStages` calculation from route's model hook to controller property, as it wasn't properly working in FastBoot mode.

app/controllers/course-overview.ts
```diff
@@ -24,6 +24,14 @@ export default class CourseOverviewController extends Controller {
     }
   }
 
+  get completedStages() {
+    if (!this.model.course) {
```
rohitpaulk, line 28: Under what circumstances would `model.course` be non-null? Wouldn't it always be a `CourseModel`?

app/controllers/course-overview.ts
```diff
@@ -24,6 +24,14 @@ export default class CourseOverviewController extends Controller {
...
+      return [];
+    }
+
+    return this.userRepositories.filter((repository) => repository.course === this.model.course).flatMap((repository) => repository.completedStages);
```
rohitpaulk, line 32:
```suggestion
    return this.userRepositories.flatMap((repository) => repository.completedStages);
```

The `userRepositories` getter already filters by course, so don't think we need the extra filter

app/routes/course-overview.ts
```diff
@@ -45,35 +45,43 @@ export default class CourseOverviewRoute extends BaseRoute {
...
-          course_id: course.id,
-        });
-      }
+  #peekCourse(slug: string) {
+    return this.store.peekAll('course').findBy('slug', slug);
+  }
 
-      return { course };
+  #queryUserRepositories(course: CourseModel) {
+    if (course && this.authenticator.isAuthenticated) {
```
rohitpaulk, line 78:
```suggestion
    if (this.authenticator.isAuthenticated) {
```

Since `course` is typed as `CourseModel`, don't think we need the extra if condition

rohitpaulk (approved):

PR #3125, opened 2025-09-25
Filter out deprecated courses for staff users in TrackController.

Arpan-206:
**Checklist**:


app/controllers/track.ts
```diff
@@ -11,7 +11,7 @@ export default class TrackController extends Controller {
 
   get courses(): CourseModel[] {
     if (this.authenticator.currentUser && this.authenticator.currentUser.isStaff) {
-      return this.model.courses;
```
rohitpaulk, line 14: Actually, let's just use the same logic as for non-staff below - the tracks page doesn't have to change at all, seems unnecessarily confusing especially with the stage counts changing

rohitpaulk (approved):

PR #3145, opened 2025-10-03
Add sortable functionality to extensions in configure extensions modal.

Arpan-206:
**Checklist**:


rohitpaulk:
- Disabled cards have a different padding distance between them than enabled cards: 

Screenshot 2025-10-16 at 11 55 21

- Disabling experience is jarring - the card initially stays in that position in the disabled state and then disappears: 

https://github.com/user-attachments/assets/bf3678ab-7ed6-4a13-8bd3-1f124d49ac6e

Ideally this'd look like a list animation, check how we handle leaderboard item movement on the CourseLeaderboard (sidebar), that's smooth

- Drag handle and text don't look vertically aligned

Screenshot 2025-10-16 at 11 57 05

Only looking at UX for now, will comment on code once this feels shippable

Issue #3187, opened 2025-10-16
Prevent light mode flash when switching from roadmap to leaderboard


rohitpaulk:
- Switching from catalog to roadmap seems fine
- Switching to the leaderboard page seems to cause issues

https://github.com/user-attachments/assets/47508323-16e0-41fd-a869-9ade30decaa2

PR #3196, opened 2025-10-20
Fix: Prevent light mode flash when switching from roadmap to lead (Is…

cwarre33:
…sue #3187)

**Checklist**:


rohitpaulk:
Clear AI slop - you've got Python code in a JS file!

PR #3245, opened 2025-11-09
chore: more natural punctuation in sqlite-"print table names"

phonty29:
The punctuation is correct, but I think this variant is more natural in continuous prose.
GitHub is little buggy when showing what has changed, so
Before:
The array size is equal to the number of cells on the page. (The page header specifies the number of cells on the page.)
After:
The array size is equal to the number of cells on the page (The page header specifies the number of cells on the page).

**Checklist**:


rohitpaulk:
Sorry we missed this! Updated at the source here: https://github.com/codecrafters-io/build-your-own-sqlite/commit/4fac86401bc66bc430e0cfaca37df15e20d253e8

Issue #3318, opened 2025-11-25
Video loads/unloads on focus toggle


rohitpaulk:
Must be a re-render triggering somewhere - lets figure out the root cause here and see if there's a way to fix.

URL: https://app.codecrafters.io/courses/shell/stages/oo8

https://github.com/user-attachments/assets/4af548c5-9bd6-4e2c-8e49-fc1ae115e7ee

PR #3400, opened 2025-12-23
Fix shell instructions video reloading using a better approach

VasylMarchuk:
Related to #3318 
Follow-up to https://github.com/codecrafters-io/frontend/pull/3366

### Brief

A better approach to fixing video reloading on "Print a prompt" Shell course stage every time `ActionCable` receives some data.

### Details

- Reverted commit ["Stop shell instructions video reloading on every ActionCable data event"](https://github.com/codecrafters-io/frontend/commit/0cd11861a9ff2366fff595bb52a0cf4aeae07e34)
- Added a new `html-safe` helper that is able to cache `SafeString` instances returned by `htmlSafe`, based on input `string`s
- Our helper now overrides the original `html-safe` helper from [`ember-cli-string-helpers`](https://github.com/adopted-ember-addons/ember-cli-string-helpers) addon
- Added some useful tests, found a simple way to actually assert `SafeString` caching

### Checklist

### DIFF
--- a/app/helpers/html-safe.ts
+++ b/app/helpers/html-safe.ts
@@ -1,0 +1,39 @@
+import Helper from '@ember/component/helper';
+import { htmlSafe, type SafeString } from '@ember/template';
+
+type Positional = [string?];
+type Return = SafeString | undefined;
+
+export interface Signature {
+  Args: {
+    Positional: Positional;
+  };
+  Return: Return;
+}
+
+/**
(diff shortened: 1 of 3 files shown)


rohitpaulk (changes requested):
Can we make this the default (override `html-safe`?). That way a dev doesn't have to actively remember to use this.

VasylMarchuk:
Ok sure, will try that

PR #3826, opened 2026-03-12
Update header logo and favicon to match landing site

sarupbanskota:
## Summary
- Replace old logomark with new logo icon SVG, with dark/light variants that swap based on color scheme (dark logo in light mode, light logo in dark mode)
- Update favicon from `.ico` to SVG with `prefers-color-scheme` media queries for light/dark mode
- Add Source Serif 4 serif font for the "CodeCrafters" logo text

## Test plan
- [ ] Verify header logo is visible on both light and dark mode
- [ ] Verify favicon adapts to system light/dark preference
- [ ] Verify "CodeCrafters" text renders in serif font

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/app/components/header/index.hbs
+++ b/app/components/header/index.hbs
@@ -32,7 +32,7 @@
     </div>
 
     {{! Desktop  }}
-    <div class="hidden md:flex items-center justify-between grow">
+    <div class="h-[56px] hidden md:flex items-center justify-between grow">
       <div class="flex items-center">
         {{! Divider }}
         <div class="bg-gray-200 dark:bg-white/5 w-px h-5 mr-3"></div>
@@ -49,7 +49,7 @@
 
           {{! Floating green bar }}
           <div
-            class="absolute -bottom-[12px] h-0.5 bg-teal-500 transition-[left,width,opacity] duration-300 ease-fling"
+            class="absolute -bottom-[18px] h-0.5 bg-teal-500 transition-[left,width,opacity] duration-300 ease-fling"
             style={{this.floatingBarStyle}}
           ></div>
         </div>


app/components/header/index.hbs
```diff
@@ -13,9 +13,10 @@
       {{! Logo }}
       <div class="flex">
         <div class="text-gray-700 dark:text-gray-300 pr-5 text-sm grow">
-          <LinkTo class="flex items-center" @route="catalog">
-            <img class="h-8 w-8 mt-3 mb-3 mr-2" src={{this.logoImage}} alt="CodeCrafters" />
-            CodeCrafters
+          <LinkTo class="flex items-center gap-[7px]" @route="catalog">
+            <img class="w-[30px] h-[21px] dark:hidden" src={{this.logoImageDark}} alt="CodeCrafters" />
```
rohitpaulk: Better to use a wordmark here so that we don't have to specify gap-[7px] etc. We can still to the hidden/block technique to get dark vs. light images. 

Also, let's only specify height, the width will be calculated automatically. This allows us to change the wordmark at any point, else it's very easy to forget to update this value

public/assets/images/favicon-black.svg
rohitpaulk: Use same terminology - favicon-dark, favicon-light instead of black/white. We also need to remove the old favicon file if it still exists

app/index.html
```diff
@@ -16,11 +16,14 @@
 
     <link integrity="" rel="stylesheet" href="{{rootURL}}assets/vendor.css">
     <link integrity="" rel="stylesheet" href="{{rootURL}}assets/codecrafters-frontend.css">
-    <link rel="icon" href="{{rootURL}}assets/favicon.ico">
+    <link rel="icon" href="{{rootURL}}assets/images/favicon-black.svg" media="(prefers-color-scheme: light)">
+    <link rel="icon" href="{{rootURL}}assets/images/favicon-white.svg" media="(prefers-color-scheme: dark)">
+    <link rel="apple-touch-icon" href="{{rootURL}}assets/images/favicon-black.svg">
```
rohitpaulk: Looks like this apple-touch-icon is new, let's move to a separate PR and find a way to test - see if we need to specify dark/light separately

rohitpaulk:
@andy1li mind taking this over please?

andy1li:
### Before vs After

**Header** is a bit shorter than before

image

---

**Footer**

image

---

**Loading** looks like this (for both light and dark themes):

image

---

**Affiliate**

image

---

**User referral**

image

---

**Institution**

image

---

**Gift card**

image

PR #3872, opened 2026-06-22
Push non-primary emails to Lobbyside (other_emails column) — take 2

TropicolX:
## Summary

Re-introduces the `other_emails` Lobbyside custom column (reverted in #3870), without the change that caused the incident.

**Root cause of the revert:** #3869 added `email-addresses` to the `include` on `GET /users/current`. That endpoint has no such serializer relationship (`EmailAddress` is an `ActiveModel` wrapper, not an AR model), so the request errored and the awaited current-user/auth bootstrap failed site-wide.

**This PR instead:**
- Loads email addresses via the dedicated `GET /api/v1/email-addresses` endpoint — the same call the institution flow already uses in prod.
- Runs it fire-and-forget + `.catch()` in `syncCurrentUser`, off the awaited path, so a failure there can never break auth again. Worst case is an empty column.
- Re-adds the pure `other_emails` derivation (excludes primary, dedupes, comma-joins) + `setVisitor` wiring + `{{did-update}}` dep.

## Verification (the part skipped last time)

- Regression test: forces `/email-addresses` to 500 and asserts the current user still loads.
- Test that `currentUser.emailAddresses` populates from the dedicated endpoint.
- 4 unit tests for the derivation.
- Live-checked the deployed endpoint on prod with a real session — returns the account's full email list for a normal user, no institution gating.
- Lint + template-lint + Glint + all 15 tests green.

## Test plan

- [x] filtered ember tests green
- [x] Confirm OTHER_EMAILS column populates on the Vercel preview before merge

Made with [Cursor](https://cursor.com)

### DIFF
--- a/app/components/lobbyside-widget/index.hbs
+++ b/app/components/lobbyside-widget/index.hbs
@@ -25,6 +25,7 @@
       this.customFields.stages_completed
       this.customFields.signed_up_date
       this.customFields.subscription_type
+      this.customFields.other_emails
     }}
     {{will-destroy this.removeScript}}
   ></div>
--- a/app/components/lobbyside-widget/index.ts
+++ b/app/components/lobbyside-widget/index.ts
@@ -15,6 +15,7 @@ declare global {
         stages_completed: string;
         signed_up_date: string;
         subscription_type: string;
+        other_emails: string;
       }): void;
     };
   }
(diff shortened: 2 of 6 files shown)


TropicolX:
bugbot run

rohitpaulk (approved):

